PR #8, opened 2016-10-12
Fix up for MSVC

borrrden:
Note the change in the StringTable::info struct

### DIFF
--- a/Fleece/Array.hh
+++ b/Fleece/Array.hh
@@ -80,6 +80,7 @@ namespace fleece {
 
         friend class Value;
         friend class Dict;
+        friend class Arr;
         template <bool WIDE> friend struct dictImpl;
     };
 
--- a/Fleece/Array.hh
+++ b/Fleece/Array.hh
@@ -80,7 +80,6 @@ namespace fleece {
 
         friend class Value;
         friend class Dict;
-        friend class Arr;
         template <bool WIDE> friend struct dictImpl;
     };
 
--- a/Fleece/MSVC_Compat.hh
+++ b/Fleece/MSVC_Compat.hh
@@ -38,7 +38,6 @@
     // WARNING: sizeof() will not work on this array since it's actually declared as a pointer.
     #define StackArray(NAME, TYPE, SIZE)    TYPE* NAME = (TYPE*)_malloca(sizeof(TYPE)*(SIZE))
 
-    #define strlcat strncat
     #define fseeko fseek
     #define ftello ftell
 
--- a/Fleece/MSVC_Compat.hh
+++ b/Fleece/MSVC_Compat.hh
@@ -38,6 +38,10 @@
     // WARNING: sizeof() will not work on this array since it's actually declared as a pointer.
     #define StackArray(NAME, TYPE, SIZE)    TYPE* NAME = (TYPE*)_malloca(sizeof(TYPE)*(SIZE))
 
+    #define strlcat strncat
+    #define fseeko fseek
+    #define ftello ftell
+
 #else
 
     #define _usuallyTrue(VAL)               __builtin_expect(VAL, true)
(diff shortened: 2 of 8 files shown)


Fleece/Array.hh
```diff
@@ -80,6 +80,7 @@ namespace fleece {
 
         friend class Value;
         friend class Dict;
+        friend class Arr;
```
snej: Actually Val and Arr aren't being used yet; they're some experimental/unfinished stuff that's checked in but shouldn't be built.

Fleece/MSVC_Compat.hh
```diff
@@ -38,6 +38,10 @@
     // WARNING: sizeof() will not work on this array since it's actually declared as a pointer.
     #define StackArray(NAME, TYPE, SIZE)    TYPE* NAME = (TYPE*)_malloca(sizeof(TYPE)*(SIZE))
 
+    #define strlcat strncat
```
snej: This is dangerous: strlcat is safer than strncat if the string overflows, so substituting strncat reopens the problem. I'd rather you found an implementation of strlcat to use.

Fleece/Value.hh
```diff
@@ -216,6 +216,7 @@ namespace fleece {
         friend class Encoder;
         friend class ValueTests;
         friend class EncoderTests;
+        friend class Arr;
```
snej: See above comment about Arr.

MSVC/Fleece/Fleece.vcxproj
```diff
@@ -25,8 +25,10 @@
     <ClCompile Include="..\..\Fleece\Fleece_C_impl.cc" />
     <ClCompile Include="..\..\Fleece\JSONConverter.cc" />
     <ClCompile Include="..\..\Fleece\KeyTree.cc" />
+    <ClCompile Include="..\..\Fleece\Path.cc" />
     <ClCompile Include="..\..\Fleece\slice.cc" />
     <ClCompile Include="..\..\Fleece\StringTable.cc" />
+    <ClCompile Include="..\..\Fleece\Val.cc" />
```
snej: Don't build Val.cc, it's not ready yet.

MSVC/Fleece/Fleece.vcxproj.filters
```diff
@@ -63,5 +63,11 @@
     <ClCompile Include="..\..\vendor\jsonsl\jsonsl.c">
       <Filter>Source Files</Filter>
     </ClCompile>
+    <ClCompile Include="..\..\Fleece\Path.cc">
+      <Filter>Source Files</Filter>
+    </ClCompile>
+    <ClCompile Include="..\..\Fleece\Val.cc">
```
snej: Ditto.

Fleece/MSVC_Compat.hh
borrrden: Hmmm I didn't know a lot about this.  However, will this still work in the presence of non-ASCII UTF-8 characters?  If not, should a function be written to handle that?

snej: It doesn't matter what the characters are; they're just uninterpreted bytes. The problem with strncat is that, while it won't overflow the destination buffer, it won't null-terminate the string if it's too long to fit. So you end up with a corrupted string. This is the official behavior of strncat and it can't be changed, so strlcat was created as a corrected version that always null-terminates the result.

borrrden: I figured out that part but wouldn't a null in the middle of a non terminated multi byte character also be corrupt later when attempting to read it as utf-8?  It probably wouldn't be as bad but it's something that worries me. That's likely to happen if the buffer can't quite fit the entire sequence and then it simply ensures the last character is null.  Is that situation recoverable?  Note that this isn't an argument in favor of strncpy, just a concern overall.

snej: All the basic `<string.h>` functions are for single-byte encodings only, including UTF-8. There's a whole separate API using a `wchar_t` type for multi-byte characters (which I've never used.)

PR #11, opened 2016-11-29
Fix utf8 windows

borrrden:
Abstract file operations into UTF-8 safe ones

### DIFF
--- a/Fleece/PlatformCompat.cc
+++ b/Fleece/PlatformCompat.cc
@@ -1,0 +1,53 @@
+#include "PlatformCompat.hh"
+#include <direct.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <wchar.h>
+#include <Windows.h>
+
+#define u8_to_wstr(in, out_name, bad_val) wchar_t out_name[256]; \
+ if(MultiByteToWideChar(CP_UTF8, 0, in, strnlen_s(in, 255) + 1, out_name, 256) == 0) errno = EINVAL; return bad_val
+
+int mkdir_u8(const char *path, int mode) {
+    u8_to_wstr(path, wname, -1);
+
+    return _wmkdir(wname);
+}
+
+int stat_u8(const char* const filename, struct stat * const s)
+{
+    u8_to_wstr(filename, wname, -1);
+    return ::_wstat64i32(wname, (struct _stat64i32 *)s);
+}
+
+int rmdir_u8(const char* const path)
+{
+    u8_to_wstr(path, wname, -1);
+    return ::_wrmdir(wname);
+}
+
+int rename_u8(const char* const oldPath, const char* const newPath)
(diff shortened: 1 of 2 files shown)


Fleece/PlatformCompat.hh
```diff
@@ -35,13 +35,19 @@
...
-    #define chmod  ::_chmod
     #define fdopen ::_fdopen
     #define fseeko fseek
     #define ftello ftell
-	#define strncasecmp _strnicmp
-	#define strcasecmp _stricmp
+    #define strncasecmp _strnicmp
+    #define strcasecmp _stricmp
+
+   int mkdir_u8(const char* const path, int mode);
```
snej: If you declare real functions here, there should be a `namespace fleece` wrapped around the contents of this header, so they don't end up in the global namespace.

Fleece/PlatformCompat.hh
```diff
@@ -57,6 +63,12 @@
 
     #define StackArray(NAME, TYPE, SIZE)    TYPE NAME[(SIZE)]
 
-    #define MKDIR(PATH, MODE) ::mkdir(PATH, (mode_t)MODE)
+    #define mkdir_u8(PATH, MODE) ::mkdir(PATH, (mode_t)MODE)
```
snej: ...and if this stuff is going into a namespace, these macros should be changed into inline functions so they can be called with `fleece::`.

Fleece/PlatformCompat.cc
```diff
@@ -0,0 +1,48 @@
+#ifdef WIN32
```
snej: The header is testing `_MSC_VER`, but here you test `WIN32`. Shouldn't they test the same thing? Otherwise if somehow we build with MSVC but not on Windows, the build will break (or something else bad will happen.)

Are there other places where we should be checking `WIN32` instead of `_MSC_VER`?

Fleece/PlatformCompat.cc
```diff
@@ -0,0 +1,48 @@
...
+
+int MIGRATE_2S(rename_u8, ::_wrename)
+
+int MIGRATE_1(unlink_u8, ::_wunlink)
+
+int MIGRATE_2(chmod_u8, ::_wchmod, int)
+
+FILE* MIGRATE_2S(fopen_u8, ::_wfopen)
+
+#endif
```
snej: Add a newline :)

Fleece/PlatformCompat.cc
borrrden: I left it like that because on windows no matter the compiler we want these definitions, but I just tried out clang and it defines `_MSC_VER` also...and for some reason doesn't define anything like `__clang__` or `__llvm__` that I can tell...so for now I'll just say don't try to use alternative compilers on Windows and change them both to `_MSC_VER`

PR #18, opened 2017-05-03
Use stdout for decoded output

avsej:
Because `usage()` banner claimed that


snej:
Thanks! I have a habit of using stderr for everything, because it's easier when debugging (stderr gets written immediately instead of being buffered) ... but in this case of course the data is the tool's regular output so it should go to stdout.

PR #19, opened 2017-05-24
Do not round up when converting from float to int number

pasin:
Removed round up so just cast.

Reference: https://github.com/couchbase/couchbase-lite-ios/issues/1764

### DIFF
--- a/Fleece/Value.cc
+++ b/Fleece/Value.cc
@@ -99,7 +99,7 @@ namespace fleece {
                 return _decLittle64(n);
             }
             case kFloatTag:
-                return (int64_t)round(asDouble());
+                return (int64_t)asDouble();
             default:
                 return 0;
         }


snej:
Why is this a bug?

pasin:
With round function, the raw value will be round up when calling asInt(). For example, if the raw value is 1.6, asInt now returns 2 instead of 1.

snej:
I think your argument is that rounding is not the same behavior as float-to-int conversions in other languages/platforms. It makes sense that Fleece should try to be compatible, but I'm not sure if the others are compatible with each other.

For C/C++ I dug up the rule:
>A finite value of any real floating type can be implicitly converted to any integer type. Except where covered by boolean conversion above, the rules are:
>* The fractional part is discarded (truncated towards zero).
>* If the resulting value can be represented by the target type, that value is used
>* otherwise, the behavior is undefined

...which is not very useful IMHO (it rounds positive numbers down but negative numbers up?)

Cocoa's NSNumber/CFNumber behaves the same way, i.e. `@(1.6).intValue == 1 && @(-1.6).intValue == -1`.

Java does the same thing:

> …the floating-point value is rounded to an integer value V, rounding toward zero using IEEE 754 round-toward-zero mode (§4.2.3)

JavaScript doesn't have integers! All numbers are floating-point.

So there seems to be consensus to round "towards zero" the way C does it.

PR #20, opened 2017-06-06
Performance: Add ratios to results

daverigby:
Add a Gross and Net ratio column to aid in comparing the relative speed of Fleece & JSON.

### DIFF
--- a/Performance.md
+++ b/Performance.md
@@ -26,13 +26,21 @@ Look up one name
 These come from the above test suites, run on a Macbook Pro with a 2.3 GHz Intel Core i7. (These are sizes and times, so smaller numbers are better.)
 
 
-|                           | JSON      | Fleece|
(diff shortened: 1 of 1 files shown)


snej:
Thanks for adding this! It makes eyeballing the numbers a little confusing, though, because the values you added are "larger is better" whereas the existing numbers are "smaller is better". What do you think of inverting your figures so the comparisons are consistent?

daverigby:
I originally had them inverted, problem was that the 28,000x number was so small (0.000035) that it seemed even more meaningless!

daverigby:
(I don't mind either way, just seemed clearer as a multiplier).

borrrden:
Unfortunately I really don't think anything is ever going to happen with this at this point, so I am going to close it.

PR #25, opened 2017-11-08
Fix compiler complaints

avsej:


avsej:
@daverigby, @snej ping

borrrden:
@snej I never noticed this PR.  @avsej What platform is giving these complaints?

avsej:
@borrrden on Fedora 26, 27. `gcc (GCC) 7.2.1 20170915 (Red Hat 7.2.1-2)`

borrrden:
We aren't planning on supporting the GCC compiler, so the caveat here will be "as long as it does not interfere with clang."  I'll find some time to see if these changes have any negative effects.

avsej:
so these changes are clang specific? enum member name really looks like used incorrectly

avsej:
> We aren't planning on supporting the GCC compiler

screenshot_2018-01-24_17-39-52

@borrrden, could you fix the README in this case, that the project does not support GCC?

Fleece/FleeceCpp.hh
```diff
@@ -249,7 +249,7 @@ namespace fleeceapi {
 
         AllocedDict(const AllocedDict &d)
         :Dict(d)
-        ,alloc_slice(d)
+        ,alloc_slice(static_cast<const fleece::alloc_slice&>(d))
```
snej: Could you change this to simply `d.data()`?

snej:
Thanks for the patch. I'd like to support GCC in Fleece, since it's a widely-useable component.

KraemerDEM:
I think this can be closed as compiling with gcc nowadays works without this patch

PR #31, opened 2018-08-04
Fix errors building with Windows

borrrden:
Still one error remains.  vasprintf is not available, and is not defined in the scope of Fleece.  Is it OK to depend on the one from LiteCore?

### DIFF
--- a/Fleece/Mutable/HeapValue.hh
+++ b/Fleece/Mutable/HeapValue.hh
@@ -14,7 +14,14 @@ namespace fleece {
         struct offsetValue {
             uint8_t _pad = 0xFF;                // Unused byte, to ensure _header is at an odd address
             uint8_t _header;                    // Value header byte (tag | tiny)
-            uint8_t _data[0];                   // Extra Value data (object is dynamically sized)
+            uint8_t _data[1];                   // Extra Value data (object is dynamically sized)
+
+#ifdef __cplusplus
+            offsetValue() {}
+        private:
+            offsetValue(const offsetValue&); // undefined
+            offsetValue& operator=(const offsetValue&); // undefined
+#endif
         };
 
         /** Stores a Value in a heap block.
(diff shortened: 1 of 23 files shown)


Fleece/Mutable/HeapValue.hh
```diff
@@ -42,8 +49,9 @@ namespace fleece {
             static void retain(const Value *v);
             static void release(const Value *v);
 
-            void* operator new(size_t size)             {return ::operator new(size);}
-
+            void* operator new(size_t size)                 {return ::operator new(size);}
+            void operator delete(void* ptr)                 {::operator delete(ptr);}
+            void operator delete(void* ptr, size_t size)    {::operator delete(ptr, size);}
```
snej: This is breaking the GCC and Clang builds (see the Travis results.) I think `operator delete` with a `size_t` parameter may be a MSVC-specific thing.

Fleece/Tree/MutableNode.hh
```diff
@@ -321,6 +325,10 @@ namespace fleece { namespace hashtree {
             return ::operator new(size + capacity*sizeof(NodeRef));
         }
 
+        static void operator delete(void* ptr, unsigned capacity) {
```
snej: This is breaking the GCC and Clang builds (see the Travis results.) I think `operator delete` with a `size_t` parameter may be a MSVC-specific thing.

Fleece/Mutable/HeapValue.hh
borrrden: I noticed that too.  Actually I think it might be a c++14 thing...I'll have to keep checking about this because I have no idea what is expected (though I have a better idea now because of the x86 Windows bug research)

borrrden:
I think this PR is finally settled.  Windows CI is up and running but most of my time was spent finding out how to make an implementation of popcnt available to MSVC on all architectures...there is no built in one.  I hope it is correct.  Are there any tests for it?

PR #41, opened 2019-01-08
JSONDelta: Fixed issues with multibyte UTF-8 chars

### LINKED ISSUE #40: Fleece deltas on UTF-8 strings produce invalid JSON strings
Fleece deltas generated on multi-byte UTF-8 strings create invalid UTF-8 sequences in the delta, which cannot be represented by JSON strings without first being encoded. This causes patching of strings to fail when the Go side consumes a delta generated by Fleece.

Given the following documents to diff, the raw bytes of the delta generated by Fleece are as follows (encoded as hexadecimal):

```
v1.json: {"mystr": "யாமறிந்த மொழிகளிலே தமிழ்மொழி போல் இனிதாவது எங்கும் காணோம், பாமரராய் விலங்குகளாய், உலகனைத்தும் இகழ்ச்சிசொலப் பான்மை கெட்டு, நாமமது தமிழரெனக் கொண்டு இங்கு வாழ்ந்திடுதல் நன்றோ? சொல்லீர்! தேமதுரத் இகழ்ச்சிசொலப் உலகமெலாம் பரவும்வகை செய்தல் வேண்டும்."}
v2.json: {"mystr": "யாமறிந்த மொழிகளிலே தமிழ்மொழி போல் இனிதாவது எங்கும் காணோம், பாமரராய் விலங்குகளாய், உலகனைத்தும் இகழ்ச்சிசொலப் பான்மை கெட்டு, நாமமது தமிழரெனக் கொண்டு இங்கு வாழ்ந்திடுதல் நன்றோ? கொண்டு! தேமதுரத் தமிழோசை உலகமெலாம் பரவும்வகை செய்தல் வேண்டும்."}
delta:   \x7B\x6D\x79\x73\x74\x72\x3A\x5B\x22\x34\x37\x38\x3D\x31\x2D\x31\x2B\x95\x7C\x35\x3D\x31\x2D\x31\x2B\xA3\x7C\x35\x3D\x31\x30\x2D\x34\x2B\x9F\xE0\xAF\x81\x7C\x32\x39\x3D\x31\x39\x2D\x34\x2B\xA4\xE0\xAE\xAE\x7C\x35\x3D\x31\x33\x2D\x31\x30\x2B\xB4\xE0\xAF\x8B\xE0\xAE\x9A\xE0\xAF\x88\x7C\x31\x30\x34\x3D\x22\x2C\x30\x2C\x32\x5D\x7D
```

We can pass the raw bytes above into [this tool as hexadecimal](https://software.hixie.ch/utilities/cgi/unicode-decoder/utf8-decoder) and see multiple "Unexpected continuation byte." appearances where the problems are occurring.

screen shot 2019-01-07 at 13 09 07

The delta works to patch the document when fed directly back into Fleece as raw bytes (e.g. in a unit test), but when unmarshalled as a UTF-8 string, the invalid UTF-8 codepoints are replaced with U+FFFD in Go, which ends up looking like this, and doesn't work for obvious reasons:

```json
{"mystr":["478=1-1+�|5=1-1+�|5=10-4+�ு|29=19-4+�ம|5=13-10+�ோசை|104=",0,2]}
```

Ideally the string diffs wouldn't be splitting bytes inside UTF-8 codepoints, and things would "just work" with valid UTF-8 codepoints, but given the string deltas are being produced by diffmatchpatch outside of Fleece, maybe we just don't do string deltas for any strings containing characters outside of the ASCII range?

This issue can also be seen in the commented out unit test inside #33

snej:
* If diff_match_patch returns a patch range that intersects a UTF-8
  multibyte character, the range is adjusted to fall on a character
  boundary. This prevents the generated JSON patch string from
  containing invalid UTF-8.
* Made gMinStringDiffLength and gTextDiffTimeout changeable.

Fixes #40

### DIFF
--- a/Fleece/Core/JSONDelta.cc
+++ b/Fleece/Core/JSONDelta.cc
@@ -37,11 +37,9 @@ namespace fleece { namespace impl {
     // (this is really just here for test purposes so we can use the JDP unit test dataset...)
     bool gCompatibleDeltas = false;
 
-    // Minimum length of strings that will be considered for diffing
-    static constexpr size_t kMinStringDiffLength = 60;
+    size_t JSONDelta::gMinStringDiffLength = 60;
 
-    // Maximum time (in seconds) that the string-diff algorithm is allowed to run
-    static constexpr float kTextDiffTimeout = 0.25;
+    float JSONDelta::gTextDiffTimeout = 0.25;
 
     // Codes that appear as the 3rd item of an array item in a diff
     enum {
@@ -49,6 +47,16 @@ namespace fleece { namespace impl {
         kTextDiffCode = 2,
         kArraymoveCode = 3,
     };
+
+
+    // Is `c` the 2nd, 3rd, ... byte of a UTF-8 multibyte character?
+    // <https://en.wikipedia.org/wiki/UTF-8#Description>
+    static inline bool isUTF8Continuation(uint8_t c) {
+        return (c & 0xc0) == 0x80;
+    }
+
+
+    static void snapToUTF8Character(long &pos, size_t &length, slice str);
 
 
 #pragma mark - CREATING DELTAS:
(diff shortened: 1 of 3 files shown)


snej:
Description of the fix, from #40:
>I fixed it by editing the patch ranges produced by diff_match_patch. For each patch that represents an insertion or deletion, the ends of the range are pushed outwards so they fall on character boundaries. This is done by looking at the first byte in the range and the first byte past the range — if either's most significant 2 bits are 10 it is a UTF-8 continuation byte, so the range needs to be pushed out past it (by decrementing its start or incrementing its end.)

snej:
I just added the test case from #40 to DeltaTests (should have done that before...) and it fails; the delta is still invalid UTF-8. Sigh. Please wait to review until I fix this.

snej:
OK, fixed now. (It was an arithmetic error in keeping track of the offset into the new string.)
I improved the tests to verify that the JSON delta is valid UTF-8, and I added some more comments to the implementation.

snej:
@borrrden That library looks great! And it's been officially moved to Github: https://github.com/nemtrif/utfcpp

PR #43, opened 2019-04-03
fix build with gcc: missed include for size_t

Dushistov:
Fix build with gcc for debug case.


borrrden:
Ok I finally added myself as a watcher on the repo so stuff like this doesn't go unnoticed!

PR #51, opened 2019-09-20
use locale independ way to convert string to double

Dushistov:
Related to couchbase/couchbase-lite-core#849 , should fix issue.
Plus this change should speedup json parsing, because of according to https://github.com/dropbox/json11/issues/38#issuecomment-214944965 when I replace in dropbox json library call strtod with google/double-conversation I see speedup in benchmark from 51ms to 40ms.

Though there is `isdigit` in vendor/jsonsl/jsonsl.c . `isdigit` locale dependent function also.
But this is related to other repository. I should note that I expect that replacing isidigit should also speedup things, because of most compilers can not inline isdigit, so check `if (ch >= '0' && ch <= '9')` is faster.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "vendor/jsonsl"]
 	path = vendor/jsonsl
 	url = https://github.com/snej/jsonsl.git
+[submodule "vendor/double-conversion"]
+	path = vendor/double-conversion
+	url = https://github.com/google/double-conversion
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,4 +35,8 @@ matrix:
 before_install:
     - eval "${MATRIX_EVAL}"
 
-script: "cd build_cmake && ./build.sh"
+script:
+  - cd build_cmake && ./build.sh
+  - cd ..
+  - LC_NUMERIC="C" ./build_cmake/FleeceTests
+  - LC_NUMERIC="ru_RU.UTF-8" ./build_cmake/FleeceTests
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,9 +27,12 @@ if(CMAKE_COMPILER_IS_GNUCC)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-psabi")
 endif()
 
+add_subdirectory(vendor/double-conversion)
 set_source_files(RESULT FLEECE_SRC)
 add_library(Fleece        SHARED  ${FLEECE_SRC})
+target_link_libraries(Fleece double-conversion)
 add_library(FleeceStatic  STATIC  ${FLEECE_SRC})
+target_link_libraries(FleeceStatic double-conversion)
 
 # "FleeceBase" static lib for clients that just need support stuff like slice, varint, RefCounted...
 set_base_platform_files(RESULT FLEECE_BASE_PLATFORM_SRC)
--- a/Fleece/Core/JSONConverter.cc
+++ b/Fleece/Core/JSONConverter.cc
@@ -18,6 +18,7 @@
 
 #include "JSONConverter.hh"
 #include "jsonsl.h"
+#include "double-conversion/string-to-double.h"
 #include <map>
 
 namespace fleece { namespace impl {
@@ -114,8 +115,13 @@ namespace fleece { namespace impl {
                 unsigned f = state->special_flags;
                 if (f & JSONSL_SPECIALf_FLOAT) {
                     char *start = (char*)&_input[state->pos_begin];
-                    char *end;
-                    double n = ::strtod(start, &end);
+		    double_conversion::StringToDoubleConverter double_conv{
+                        double_conversion::StringToDoubleConverter::ALLOW_TRAILING_JUNK, 0., 0., "Infinity", "NaN"
+                    };
+		    int processed_characters_count = 0;
+		    double n = double_conv.StringToDouble(start,
+						      static_cast<int>(_input.size - state->pos_begin),
+						      &processed_characters_count);
                     _encoder.writeDouble(n);
                 } else if (f & JSONSL_SPECIALf_UNSIGNED) {
                     _encoder.writeUInt(state->nelem);
--- a/Fleece/Core/Value.cc
+++ b/Fleece/Core/Value.cc
@@ -31,6 +31,7 @@
 #include "ParseDate.hh"
 #include <math.h>
 #include "betterassert.hh"
+#include "double-conversion/double-to-string.h"
 
 
 namespace fleece { namespace impl {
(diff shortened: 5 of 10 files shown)


borrrden:
I hesitate to pull in an entire new dependency just for this one line.  If JSON requires the dot to be used, then stringstreams could be used with an imbued locale in order to ignore the global one.  What do you think of that approach?

Dushistov:
And I find out that this is not final fix, 
I got "17,42342" in json after document retrieve from database.
I suspect that this is because of:

```
            case kFloatTag: {
                if (_byte[0] & 0x8)
                    sprintf(str, "%.16g", asDouble());
                else
                    sprintf(str, "%.6g", asFloat());
                break;
            }
```

yeah, `sprintf` depend on locale also.

Dushistov:
@borrrden 

I fix also sprintf case.

Not that after reading code I can say that I not only fix locale issue,
but potentially also fix information loss.

The code before this patch used `%.6g` and `%.16g`, that works for most cases,
but not for all.
For example:

```
char buf[32];
float origin_f = 1.0f / 3.0f;
sprintf(s, "%.6g", x);
float new_f = strtod(buf, nullptr);
```

after this `new_f != origin_f`, so during save to database, load from database you can
loose bits of information in your number, for this particular case you need `%.8g`.
Google's ToShortest function take care about this case. So after that patch store/load from database should not cause information lost.

borrrden:
`c4db_encodeJson` is actually not used anywhere in the library, so no need to optimize it (it is a convenience function that Couchbase Lite does not need, but may be useful to others and we use it in our tests too).  I have been doing some experimentation and I found a locale specific `sprintf` and `strtod` and even used it before I realized that it's only available on certain platforms (BSD and Windows....).  I'm leaning toward an implementation that can somehow push and pop the locale setting though, rather than a full on 3rd party library.  The loss of information from float to string is an acceptable one and our advise has been "if you need that level of precision then use strings instead of numbers."

borrrden:
Let me know what you think of #53

Dushistov:
> c4db_encodeJson is actually not used anywhere in the library,

I find out about this function from reading something like native_c4database.cc from couchbase-lite-java , and I used it a lot in  in my [Rust wrapper around couchbase-lite-core](https://github.com/dushistov/couchbase-lite-core). Though have no idea that this is not frequently used function, but I don't see alternative for saving JSON directly.

> The loss of information from float to string is an acceptable one and our advice has been "if you need > that level of precision then use strings instead of numbers."

But why? According to google developers 17 characters enough to encode any number (see kBase10MaximalLength constant in double-conversation code).
Why loose any information, if it is possible to not loose anything? 

And it is really inconvenient, you see JSON encoding/decoding function is generated from language types, obviously I can mark any field with float/double to use special function that encode/decode them to stirng, but it is a lot of manual work, and what if I forget to mark field as such?
The more robust solution will be patch fleece every time and do not care about such issue in the main application.

Plus I want to share this data with somebody who use other language (frontend), it will be inconvenient to document every field like - it is really number, but I save it as string,
it is really, really strange.

Dushistov:
@borrrden

> Let me know what you think of #53

Access to thread-local variables in shared-library is not cheap.
And you call it every sprintf/strtod call (I mean uselocale). It may potentially slow down all program.

borrrden:
Well there is no alternative for saving JSON directly because that's not what we do in Couchbase Lite.  We turn domain objects (C# classes, etc) into Fleece by using `FLEncoder`.  Thus we now have hundreds of lines of 3rd party code added for something that doesn't help the Couchbase Lite distributed product.  We based the 6 and 16 values off of information about how many accurate bytes there are in each type (their precision).  

I don't remember hearing about the Rust wrapper before, that's pretty neat!  

@snej What are your thoughts on this?

snej:
> there is no alternative for saving JSON directly because that's not what we do in Couchbase Lite.

The replicator encodes documents into JSON when pushing. So JSON-generation issues are definitely important to Couchbase Lite.

snej:
What about using `strtod_l`, which explicitly takes a `locale_t` parameter?

`<xlocale.h>` defines a macro `LC_C_LOCALE` that corresponds to the C system locale, but I'm not sure that's cross-platform (might be one of the BSD-isms in macOS X.)

snej:
Found this in [a post](http://lua-users.org/lists/lua-l/2016-04/msg00216.html) on the lua mailing list:

>   - strtod_l is available on linux (at least both glibc and musl) if _GNU_SOURCE is defined
>   - _strtod_l is available in MSVC since VS2005
>   - strtod_l is available on OSX since Darwin 8.
>   - strtod_l is available on FreeBSD since 9.1

I'm working on a fix using that.

snej:
> after this new_f != origin_f, so during save to database, load from database you can
loose bits of information in your number

Comparing floating point numbers for equality before and after some transformation is always a problem, because of roundoff. I've had to explain to testers many times that they can't write assertions that compare floats for equality.

This is especially bad with 32-bit floats. I don't think anyone who cares about accuracy is going to choose those over 64-bit, when you get fewer than 7 significant [decimal] digits.

I came up with the `%.6g` and `%.16g` formats after some experimentation, to get the best accuracy without getting a lot of badly-rounded strings for short decimal values. For example, 0.9f formatted to 8 figures is `0.89999998`. We got several bug reports about this before I tweaked the formatting! I just did some tests, formatting all multiples of 0.001f from 0 to 1, and 7 figures seems to work well; I'm not sure why I went down to 6 originally, but I must have found some cases that formatted badly.

Dushistov:
@snej 

> Comparing floating point numbers for equality before and after some transformation is always a problem,

In this particular case I suppose that comparing is the only and right way to test equality. 

You see one thing that you calculate something, obviously `float` and `double` give you some round error. Another thing is serialization / deserialization. 

For example you have JSON document with two properties `A` and `B`, 
`A` has float point number type.
You change `B`, save and reload JSON document from database.
Repeat it several times. What if after several such transformation after with bits representation of float/double was changes in the way that error would be accumulated?

If you store and load the same bits representation you can be sure that you have no problem at all. I am not sure that after several cycles for save/load `.6g` and `.16g` the error will be not accumulated

borrrden:
Closing in favor of #56 (Thanks for getting us to address the issue though!)

snej:
> Though there is isdigit in vendor/jsonsl/jsonsl.c . isdigit locale dependent function also.

Actually it isn't; I just checked the [C reference docs](https://en.cppreference.com/w/c/string/byte/isdigit): "isdigit and isxdigit are the only standard narrow character classification functions that are not affected by the currently installed C locale."

Although you may be referring to this: "...although some implementations (e.g. Microsoft in 1252 codepage) may classify additional single-byte characters as digits."

I checked `<ctype.h>` in the macOS/iOS SDKs, and the implementation isn't locale-dependent but it looks less efficient than it could be — it's using table lookup.

snej:
>I expect that replacing isdigit [in jsonsl] should also speedup things

I tried this yesterday [on macOS] and the speedup was very small. I guess Clang's (Apple's?) `isdigit` is inlined.

PR #52, opened 2019-09-21
Make string parsing locale independent

borrrden:


borrrden:
Nevermind, apparently this is only a BSD / Windows thing and the world cannot have nice things...

PR #53, opened 2019-09-21
CBL-415 Create locale independent versions of sprintf and strtod

borrrden:
Make use of per-thread locales


borrrden:
Closed in favor of #56

PR #56, opened 2019-09-23
Fixed issues with reading/writing JSON floats

### LINKED ISSUE #54: Errors parsing & writing JSON floats, in some locales
The code we use for reading/writing floating point is (accidentally) locale-dependent, thanks to some _really_ bad design decisions in the ANSI C and POSIX specs. This means that in locales that don't use "." as the decimal point character, floats won't parse or format properly. (Pointed out by @Dushistov in #51.)

### LINKED ISSUE #55: Roundoff error in round-trip of floats from Fleece->JSON->Fleece
The way Fleece formats floating point numbers when writing JSON ensures they don't end up with long repeating decimals (like "0.8999999999" instead of "0.9") but can cause a small amount of round-off error. It's been suggested that we can do better than this.

snej:
* Reading/writing floating point was accidentally locale-dependent,
  which caused errors in locales that don't use '.' as the decimal
  point. (Fixes #54)
* Improved round-trip accuracy of Fleece->JSON->Fleece float/double
  conversions, by writing the optimal number of decimal places.
  To do this we brought in a small subcomponent of Swift. (Fixes #55)

This PR does the same as #51, basically, but brings in a lot less 3rd party code.


Dushistov:
One thing that is bad, that in compare with google/double-conversation there are no included unit tests that you can run on every platform and every compiler, it should be fine for Mac OS/iOS,
but what about other? Almost nobody uses swift on Linux/Windows/Android even if there is port to that platform.

snej:
> there are no included unit tests that you can run on every platform and every compiler, 

True. But I still chose to use this library over the larger Google one.

>it should be fine for Mac OS/iOS,
>but what about other? Almost nobody uses swift on Linux/Windows/Android even if there is port to that platform.

Not really. There's a nontrivial Swift community on Linux; IBM is a big backer of server-side Swift. And Swift itself is well-tested.

snej:
It's alllllmost building on Windows. MSVC is just complaining about the literal `0x1.0p24F`, which I have to admin confuses me too. It looks like this is a hex representation of a float‽

snej:
... yup, it's a hexadecimal floating-point literal, and the `p` is the binary equivalent of `e` in a regular float. And my suspicion was correct, this is a [C99 feature](https://stackoverflow.com/questions/4825824/hexadecimal-floating-constant-in-c) which is why MSVC doesn't like it.

The good news is that this syntax was [added to C++17](https://en.cppreference.com/w/cpp/language/floating_literal) so we can work around the problem by bumping the compiler dialect to C++17 ... see, @borrrden, I found a good reason to switch! 🤣

snej:
Oh wonderful — Clang 5 claims to support C++17 (`__cplusplus >= 201700L` is true) but doesn't have `string_view`.

snej:
Apparently MSVC doesn't support hex float literals at all. There's no mention of them in [their docs](https://docs.microsoft.com/en-us/cpp/cpp/numeric-boolean-and-pointer-literals-cpp?view=vs-2019).

snej:
According to [someone on SO](https://stackoverflow.com/a/44578923/98077), "Visual C++ will have [hex float literals] available in Visual Studio 2017 15.6 release."
Whereas the logs say our Appveyor build uses Visual Studio 14.0. :(

snej:
I backed out the C++17 stuff and just changed the hex FP literals to regular ones.
Also added a subset of the Swift unit tests for the formatter.

borrrden:
Yes, here is the official blog that says the feature is available in 15.5 (aka VS 2017).  We can change AppVeyor to use that, but we also need to change our entire build infrastructure to move from VS2015 to VS2017 (something that eventually we will need to do)

PR #57, opened 2019-09-30
fix misside xlocale.h issue

Dushistov:
locale.h is conforming POSIX.1‐2008, while xlocale.h is not standard heder and was removed from glibc since version 2.26:

https://sourceware.org/ml/libc-alpha/2017-08/msg00010.html

* The nonstandard header <xlocale.h> has been removed.  Most programs should
  use <locale.h> instead.  If you have a specific need for the definition of
  locale_t with no other declarations, please contact
  libc-alpha@sourceware.org and explain.

So build failed on all recent Linux ditros.

### DIFF
--- a/Fleece/Support/NumConversion.cc
+++ b/Fleece/Support/NumConversion.cc
@@ -20,7 +20,7 @@
 #include "SwiftDtoa.h"
 #include <locale.h>
 #include <stdlib.h>
-#ifndef _MSC_VER
+#if !defined(_MSC_VER) && !defined(__GLIBC__)
 #include <xlocale.h>
 #endif
 


borrrden:
Is there any other way to fix this?  As you said it is not removed until 2.26 and the version we build against is currently 2.17 meaning that we'd want to include it there.  Is there a numeric define for GLIBC?

Dushistov:
@borrrden 

Sorry, I forgot to mention it. Before glibc 2.26 things looks like:

```
//locale.h
#include <xlocale.h>
```

so this patch is no harm for any version prior 2.26 and fix compilation after 2.26.

So on Linux/glibc it is useless include xlocale.h in any case, except you want speedup compilation.

borrrden:
Oh, fair enough.  Looks like macOS is the odd one out then.

PR #60, opened 2019-11-05
Fix issue 59

### LINKED ISSUE #59: converting scientific numbers without decimal separator
When you try to convert a json to fleece containing a scientific number without a decimal separator only the part before "e" is converted: E. g. `4e04 --> 4`

Simmay93:
Fixes #59


borrrden:
@snej This one seems fine right?

snej:
Thanks for the fix!

PR #61, opened 2019-11-15
CBL-540: fix: memory leak with platform string

jayahariv:
* include destructor specific to platform, so that CFString can be released

### DIFF
--- a/Fleece/Core/SharedKeys.cc
+++ b/Fleece/Core/SharedKeys.cc
@@ -21,6 +21,14 @@
 #include "FleeceException.hh"
 
 
+#ifdef __APPLE__
+#include <CoreFoundation/CFString.h>
+typedef CFStringRef PlatformString;
+#else
+typedef void* PlatformString;
+#endif
+
+
 #define LOCK(MUTEX)     lock_guard<mutex> _lock(MUTEX)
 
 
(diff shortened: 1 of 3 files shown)


snej:
This code will work, but it's made the class inconsistent. Formerly it knew nothing about what a `PlatformString` is; it just treats it as an opaque pointer. Now it knows that (on Apple platforms) it's a `CFStringRef`.

If it knows that, and releases the strings on destruct, then to be consistent **it should retain them when they're added** too. 

And we can **make the type meaningful**, i.e.

```c
#ifdef __APPLE__
typedef CFStringRef PlatformString;
#else
typedef void* PlatformString;
#endif
```

Could you make these changes?

borrrden:
If you do that then I think you will need to add another release somewhere else or risk coming back to the same problem again (i.e. after adding them, release them from the caller who added them)

jayahariv:
> If you do that then I think you will need to add another release somewhere else or risk coming back to the same problem again (i.e. after adding them, release them from the caller who added them)

I have removed the retain when setting the key by caller, and include inside the setPlatformMethod.

Fleece/Core/SharedKeys.cc
```diff
@@ -21,13 +21,30 @@
 #include "FleeceException.hh"
 
 
+#ifdef __APPLE__
```
snej: This should be in the header, replacing the existing typedef of PlatformString. Here you've created a _second_ typedef of PlatformString, this time outside any namespace, which isn't being used by anything.

snej: Note that you will have split the `#include` out into a separate `#ifdef` block at the top of the header, since you don't want to include a header while inside a namespace or class definition...

jayahariv: done!

PR #65, opened 2019-11-26
Upgrade Catch

borrrden:
Catch1 is not doing so well with C++17 on Windows.

### DIFF
--- a/API/fleece/FLSlice.h
+++ b/API/fleece/FLSlice.h
@@ -86,7 +86,7 @@ static inline FLSlice FLStr(const char *str) {
 // Macro version of FLStr, for use in initializing compile-time constants.
 // STR must be a C string literal. Has zero runtime overhead.
 #ifdef _MSC_VER
-    #define FLSTR(STR) {("" STR), sizeof(("" STR))-1}
+    #define FLSTR(STR) (FLSlice({("" STR), sizeof(("" STR))-1}))
 #else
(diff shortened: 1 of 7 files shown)


borrrden:
Clang 5 is failing because of some C++17 stuff, oddly.  I'll have to look into how the `-std` flag interacts with the `__cplusplus` define to see which path forward to take.

borrrden:
Hold off on this, there are so many things that I have no idea about how to upgrade to Catch2 (NWayTestCase ?  )

API/fleece/FLSlice.h
```diff
@@ -86,7 +86,7 @@ static inline FLSlice FLStr(const char *str) {
 // Macro version of FLStr, for use in initializing compile-time constants.
 // STR must be a C string literal. Has zero runtime overhead.
 #ifdef _MSC_VER
-    #define FLSTR(STR) {("" STR), sizeof(("" STR))-1}
+    #define FLSTR(STR) (FLSlice({("" STR), sizeof(("" STR))-1}))
```
snej: Are you sure this is valid C?

borrrden: It's not....ugh....but the original is not great C++ either and causes rampant errors when used inline :(

snej:
This looks reasonable for Fleece. Have you verified that LiteCore's tests build and pass? (Besides the Fleece header changes, LiteCore tests use the same copy of catch.hpp in the Fleece repo.)

PR #66, opened 2020-01-14
Safety fixes for Encoder::writeKey(Value*)

borrrden:
Ensure it won't write an integer key belonging to a different
SharedKeys, which would result in the wrong key string.
(This is a speculative fix, since I haven't seen this occur, but I
have suspicions this might be behind CBSE-7789.)


snej:
Isn't this just a cherry-pick of my commit 782c83afc0db034a6d3c9de52d4d4be0a2fb7aca? Why do it as a pull request?

PR #79, opened 2020-10-13
Fix ARM64 MVS build

ionutcosminmihaipfa:
The build on MVS 16 2019 for Debug|ARM64 was not working.

### DIFF
--- a/Fleece/Support/endianness.h
+++ b/Fleece/Support/endianness.h
@@ -66,7 +66,7 @@
      defined(__ARMEL__) || defined(__THUMBEL__) || defined(__AARCH64EL__) || \
      defined(_MIPSEL) || defined(__MIPSEL) || defined(__MIPSEL__) || \
      defined(_M_IX86) || defined(_M_X64) || defined(_M_IA64) || /* msvc for intel processors */ \
-     defined(_M_ARM) /* msvc code on arm executes in little endian mode */
+     defined(_M_ARM) ||  defined(_M_ARM64) /* msvc code on arm executes in little endian mode */
 #    define __LITTLE_ENDIAN__
 #  endif
 #endif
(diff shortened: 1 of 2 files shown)


Fleece/Support/endianness.h
```diff
@@ -130,7 +129,7 @@
 #  define ntoh64(x)     (x)
 #  define hton64(x)     (x)
 #  else
-#    warning "UNKNOWN Platform / endianness; network / host byte swaps not defined."
+#   pragma message ( "UNKNOWN Platform / endianness; network / host byte swaps not defined.")
```
borrrden, line 132: What is the purpose of changing this?

ionutcosminmihaipfa: MSVC doesn't support #warning directive, it issues a `C1021: invalid preprocessor command 'warning'` error. #pragma message is supported by MSVC, clang, gcc compilers.

borrrden: Wow so it is!  I swore it worked in MSVC too but in fact it doesn't.  Let's add "warning:" to the beginning of the message.  I find that just emitting a message won't show anything in the errors / warnings tab of VS but adding "warning: " to the beginning will make it into a warning.  That's very unfortunate since now on both compilers you will get an entry with no line number or any sort of information other than maybe the file it was in....this seems like a rare enough case though so

borrrden:
We don't officially support ARM64 for Windows yet, but these changes are small enough not to cause a bother but are these all the changes that are needed or are there many more in other repos?

ionutcosminmihaipfa:
I could compile couchbase-lite-C for ARM64 just with these changes, but I didn't run the app. I guess making the project to compile for arm64 is a step forward.

Fleece/Support/endianness.h
ionutcosminmihaipfa: Done

ionutcosminmihaipfa:
Any observations left for this issue?

borrrden (approved):

borrrden:
Nope, looks alright.  Thanks!

PR #80, opened 2021-01-11
make catalyst compatible

jayahariv:
* SDKROOT = iphoneos to make it compatible with catalyst


jayahariv:
I assume, travis failure is not related with this PR.

borrrden (approved):

snej:
I had to back this out because it breaks the unit-tests target in Xcode — the iOS SDK doesn't support building command-line tools.

PR #81, opened 2021-01-13
Add in a method to zero the memory of a slice

borrrden:


API/fleece/slice.hh
```diff
@@ -238,7 +238,8 @@ namespace fleece {
 
         void setBuf(const void *b NONNULL) noexcept          {pure_slice::setBuf(b);}
         void setSize(size_t s) noexcept                      {pure_slice::setSize(s);}
-        void shorten(size_t s)                      {assert_precondition(s <= size); setSize(s);}
+        void shorten(size_t s)                               {assert_precondition(s <= size); setSize(s);}
+        void clear()                                         { memset((void *)buf, 0, size); }
```
snej, line 242: The name `clear` doesn't seem right to me. STL containers have `clear` methods that reset them to zero size, and LiteCore has some methods like that too. How about `wipe` or `zeroFill`?

API/fleece/Fleece.hh
```diff
@@ -578,6 +578,8 @@ namespace fleece {
         const alloc_slice& data() const                 {return *this;}
         explicit operator bool () const                 {return Dict::operator bool();}
 
+        void clear() { alloc_slice::clear(); }
```
snej, line 581: Nit: Make the indentation of the method body consistent with nearby lines

API/fleece/slice.hh
```diff
@@ -350,6 +351,7 @@ namespace fleece {
         void reset(size_t);
         void resize(size_t newSize);
         void append(pure_slice);
+        void clear() { memset((void *)buf, 0, size); }
```
snej, line 354: Nit: Make the indentation of the method body consistent with nearby lines

snej (changes requested):

davemilter:
Sorry for the interference,
but if this is for security, then `memset` or loop should not be used:
 https://rules.sonarsource.com/c/type/Vulnerability/RSPEC-5798

borrrden:
This intrusion is most welcome.  I never would have known about this otherwise....

PR #86, opened 2021-03-09
handle json encoder to handle ObjC types

jayahariv:
* handle json encoder to handle ObjC types
* combined calls to a single function `encodeTo:`

### DIFF
--- a/ObjC/Encoder+ObjC.mm
+++ b/ObjC/Encoder+ObjC.mm
@@ -25,7 +25,7 @@ using namespace fleece::impl;
 
 
 @interface NSObject (FleeceInternal)
-- (void) fl_encodeTo: (FLEncoderImpl*)enc;
+- (void) fl_encodeToFLEncoder: (FLEncoder)enc;
 @end
 
 
@@ -62,13 +62,13 @@ bool FLEncoder_WriteNSObject(FLEncoder encoder, id obj) FLAPI {
 
 @implementation NSObject (Fleece)
 - (void) fl_encodeToFLEncoder: (FLEncoder)enc {
-    [self fl_encodeTo: enc];
+    [self fl_encodeToFLEncoder: enc];
 }
 @end
 
 
 @implementation NSObject (FleeceInternal)
-- (void) fl_encodeTo: (FLEncoderImpl*)enc {
+- (void) fl_encodeToFLEncoder: (FLEncoder)enc {
     // Default implementation -- object doesn't implement Fleece encoding at all.
     NSString* msg = [NSString stringWithFormat: @"Objects of class %@ cannot be encoded",
                      [self class]];
@@ -79,14 +79,14 @@ bool FLEncoder_WriteNSObject(FLEncoder encoder, id obj) FLAPI {
 
 
 @implementation NSNull (Fleece)
-- (void) fl_encodeTo: (FLEncoderImpl*)enc {
+- (void) fl_encodeToFLEncoder: (FLEncoder)enc {
     ENCODER_DO(enc, writeNull());
 }
 
 @end
 
 @implementation NSNumber (Fleece)
-- (void) fl_encodeTo: (FLEncoderImpl*)enc {
+- (void) fl_encodeToFLEncoder: (FLEncoder)enc {
     switch (self.objCType[0]) {
         case 'b':
             ENCODER_DO(enc, writeBool(self.boolValue));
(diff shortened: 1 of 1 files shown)


pasin (approved):
Can you match the code style (e.g. space after :, pointer * with no front space, and nospace after (type)? I think it's better to preserve fl_ prefix. Apart from that, the PR looks good to me.

ObjC/Encoder+ObjC.mm
```diff
@@ -25,7 +25,7 @@
 
 
 @interface NSObject (FleeceInternal)
-- (void) fl_encodeTo: (Encoder*)enc;
+- (void) fl_encodeTo: (FLEncoderImpl*)enc;
```
snej, line 28: `FLEncoderImpl*` is literally identical to `FLEncoder` (there's a typedef at Fleece+ImplGlue.hh:44), so this makes `fl_encodeTo:` identical to `fl_encodeToFLEncoder:`.

 it would be simpler just to change it to `fl_encodeToFLEncoder:` in this source file, so there's only one method.

snej (changes requested):

ObjC/Encoder+ObjC.mm
```diff
@@ -25,7 +25,7 @@
 
 
 @interface NSObject (FleeceInternal)
-- (void) fl_encodeTo: (Encoder*)enc;
+- (void) fl_encodeToFLEncoder: (FLEncoder)enc;
```
snej, line 28: This was already declared in Fleece+CoreFoundation.h. You can just remove this`@interface`.

ObjC/Encoder+ObjC.mm
```diff
@@ -59,113 +59,101 @@ bool FLEncoder_WriteNSObject(FLEncoder encoder, id obj) FLAPI {
 }
 
 
+
 @implementation NSObject (Fleece)
 - (void) fl_encodeToFLEncoder: (FLEncoder)enc {
-    // Fall back to the internal fl_encodeTo:, which takes a raw C++ Encoder*.
-    [self fl_encodeTo: ((FLEncoderImpl*)enc)->fleeceEncoder.get()];
+    [self fl_encodeToFLEncoder: enc];
```
snej, line 65: This is the same method name as declared right below, and also it calls itself so it would crash ... just remove this `@implementation`.

ObjC/Encoder+ObjC.mm
```diff
@@ -59,113 +59,101 @@ bool FLEncoder_WriteNSObject(FLEncoder encoder, id obj) FLAPI {
...
 }
 @end
 
 
 @implementation NSObject (FleeceInternal)
-- (void) fl_encodeTo: (Encoder*)enc {
+- (void) fl_encodeToFLEncoder: (FLEncoder)enc {
```
snej, line 71: The category name is `Fleece`, not `FleeceInternal`.

ObjC/Encoder+ObjC.mm
```diff
@@ -59,113 +59,101 @@ bool FLEncoder_WriteNSObject(FLEncoder encoder, id obj) FLAPI {
...
 }
+
 @end
 
 
 @implementation NSNull (Fleece)
-- (void) fl_encodeTo: (Encoder*)enc {
-    enc->writeNull();
+- (void) fl_encodeToFLEncoder: (FLEncoder)enc {
+    ENCODER_DO(enc, writeNull());
```
snej, line 83: It's probably better to just call `FLEncoder_WriteNull`, etc. than use the internal `ENCODER_DO` in all these methods.

ObjC/Encoder+ObjC.mm
```diff
@@ -59,113 +59,101 @@ bool FLEncoder_WriteNSObject(FLEncoder encoder, id obj) FLAPI {
...
+    ENCODER_DO(enc, writeData(slice(self)));
 }
+
 @end
 
 @implementation NSArray (Fleece)
 - (void) fl_encodeToFLEncoder: (FLEncoder)enc {
-    FLEncoder_BeginArray(enc, (uint32_t)self.count);
+    ENCODER_DO(enc, beginArray((uint32_t)self.count));
     for (NSString* item in self) {
```
snej, line 139: I just realized that this should be `id item`, not `NSString*`.

snej (changes requested):

snej (approved):
Great!

PR #87, opened 2021-03-11
expose isFleece API

jayahariv:
* this will enable us to include content or not, when encoding blob. 
* Blob requires to encode content, when encode as Query params(Fleece encoder), where as when encode as JSON string, it didn't require the content(JSON Encoder).
* usage can be found in the linked lite-ios PR (couchbase/couchbase-lite-ios#2812)


pasin (commented):
I'm not sure if this is needed as it's used to branch the encoding logic in an indirect way.

jayahariv:
Closing, since we are moving in different direction.

PR #89, opened 2021-03-18
Remove includes of private headers from public ones

borrrden:
This required some shifting around of a few things like hot, cold, and stepover.


borrrden:
Windows is hanging because the fix has not been taken from master yet...

jianminzhao (approved):

PR #92, opened 2021-03-25
fix setInt(INT i) when INT=int64_t

jianminzhao:


borrrden:
This is the danger of requesting review before validation completes.  Linux is a stickler with its type sizes that are different than mac or Windows.  Let me know when it is fixed.

snej (approved):

PR #93, opened 2021-03-31
API: Added FLSlot_SetArray, FLSlot_SetDict

snej:
Simple inlines to avoid having to cast to (FLValue)

### DIFF
--- a/API/fleece/Fleece.h
+++ b/API/fleece/Fleece.h
@@ -368,6 +368,14 @@ extern "C" {
     void FLSlot_SetData(FLSlot FLNONNULL, FLSlice) FLAPI;    ///< Stores a data blob into a slot.
     void FLSlot_SetValue(FLSlot FLNONNULL, FLValue) FLAPI;   ///< Stores an FLValue into a slot.
 
+    static inline void FLSlot_SetArray(FLSlot NONNULL slot, FLArray array) {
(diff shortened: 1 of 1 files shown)


API/fleece/Fleece.h
```diff
@@ -368,6 +368,14 @@ extern "C" {
     void FLSlot_SetData(FLSlot FLNONNULL, FLSlice) FLAPI;    ///< Stores a data blob into a slot.
     void FLSlot_SetValue(FLSlot FLNONNULL, FLValue) FLAPI;   ///< Stores an FLValue into a slot.
 
+    static inline void FLSlot_SetArray(FLSlot NONNULL slot, FLArray array) {
```
borrrden, line 371: I wonder if array should be non-null.  Would a null array be a shortcut for `SetNull` ?

borrrden (changes requested):
Could we have the mutable variants too? :)

borrrden:
Does #95 supersede this?

snej:
> Could we have the mutable variants too? :)

`FLMutableArray` is just a non-const `FLArray`, so they're type compatible already.

snej:
> Does #95 supersede this?

No, it just means more inlines for me to write...

borrrden (approved):

PR #94, opened 2021-04-06
CBL-1786: Don't ignore unknown warnings with clang

borrrden:
clang will throw a warning about an unrecognized warning and MSVC doesn't use this at all


bmeike (approved):
These warnings are annoying.  I hope we can make them go away soon...

PR #96, opened 2021-04-22
Fix a reference-leak bug, #CBL-1707

jianminzhao:
Preserving the invariant in MDict: _newKeys holds all alloc_slices for the keys of _map.

### DIFF
--- a/Fleece/Integration/MDict.hh
+++ b/Fleece/Integration/MDict.hh
@@ -64,6 +64,7 @@ namespace fleece {
             MCollection::initAsCopyOf(d, isMutable);
             _dict = d._dict;
             _map = d._map;
+            _newKeys = d._newKeys;
             _count = d._count;
         }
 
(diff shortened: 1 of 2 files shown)


borrrden (commented):
Unfamiliar with this area actually, since it is only used by iOS.

pasin (approved):

Fleece/Integration/MDict.hh
```diff
@@ -137,8 +138,9 @@ namespace fleece {
                 return true;
             MCollection::mutate();
             _map.clear();
+            _newKeys.clear();
             for (Dict::iterator i(_dict); i; ++i)
-                _map.emplace(i.keyString(), MValue::empty);
```
snej, line 141: This change isn't necessary. `i.keyString()` is owned by `_dict` so we don't need to make our own copy of it.

snej (changes requested):

Fleece/Integration/MDict.hh
jianminzhao: I try to maintain the invariant that: all keys, as slices, of _map are allocated in alloc_slices held in _newKeys. And I am afraid this is actually the critical part of this commit. I'll double check it.

PR #98, opened 2021-04-30
API: Added printf-style formatted builder functions

snej:
These functions let you create complex nested Fleece structures with
a single call. It's basically JSON with values substituted from
arguments using printf-style "%" specs. You can
* Create a mutable array or dict
* Add to an existing mutable array or dict
* Write to an encoder

See doc-comments in Fleece.h for details. BuilderTests.cc has a bunch of examples.

----
#### API: Added [Core]Foundation-to-Fleece conversions
Unlike the existing `FLEncoder_WriteCFObject`, these functions create
mutable object trees. They're needed by the formatted builder so it
can accept complex CF/NS values using the `%@` specifier.

### DIFF
--- a/API/fleece/Fleece+CoreFoundation.h
+++ b/API/fleece/Fleece+CoreFoundation.h
@@ -36,6 +36,16 @@ extern "C" {
         Caller must CFRelease the result. */
     NODISCARD FLEECE_PUBLIC CFTypeRef FLValue_CopyCFObject(FLValue FL_NULLABLE) FLAPI;
 
+    /** Copies a CoreFoundation object to a standalone Fleece object, if possible.
+        @throws an NSException if the object is not of a class that can be converted, or if it
+                contains such an object.
+        \note  You must call \ref FLValue_Release when finished with the result. */
+    FLValue FLValue_FromCFValue(CFTypeRef);
+
+    /** Stores a CoreFoundation object into a slot in a Fleece dict/array, if possible.
+        This is supported for the CF/NS equivalents of Fleece types: CFString, CFNumber,
+        CFData, CFArray, CFDictionary. */
+    void FLSlot_SetCFValue(FLSlot slot, CFTypeRef value);
 
     /** Same as FLDictGet, but takes the key as a CFStringRef. */
     NODISCARD FLEECE_PUBLIC FLValue FLDict_GetWithCFString(FLDict FL_NULLABLE, CFStringRef) FLAPI;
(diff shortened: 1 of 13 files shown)


Fleece/Core/Builder.cc
```diff
@@ -0,0 +1,619 @@
...
+        string readLiteralString() {
+            string out;
+            const char quote = get();               // single or double-quote
+            char c;
+            while (quote != (c = get())) {
+                if (c == '\\') {
+                    switch ((c = get())) {
+                        case 'n':   c = '\n'; break;
+                        case 'r':   c = '\r'; break;
+                        case 't':   c = '\n'; break;
```
pasin, line 414: \t ?

API/fleece/Fleece.h
```diff
@@ -1238,6 +1239,67 @@ while (NULL != (value = FLDictIterator_GetValue(&iter))) {
...
+         @{
+        These functions use the `printf` idiom to make it convenient to create structured Fleece
+        values in memory with one call. They create or modify a `FLMutableArray` or `FLMutableDict
+        by reading the given format string and the following arguments.
+
+        The format string is basically JSON5, except that any value in it may be a printf-style
+        '%' specifier instead of a literal, in which case that value will be read from the next
+        argument. The supported format specifiers are:
+
+        - Boolean:           `%c` (cast the arg to `char` to avoid a compiler warning)
```
pasin, line 1255: or '%b'?

API/fleece/Fleece+CoreFoundation.h
```diff
@@ -68,6 +78,11 @@ extern "C" {
     /** Same as FLEncoder_Finish, but returns result as NSData or error as NSError. */
     NSData* FLEncoder_FinishWithNSData(FLEncoder, NSError** FL_NULLABLE) FLAPI;
 
+    /** Copies a Foundation object to a standalone Fleece object, if possible.
+        @throws an NSException if the object is not of a class that can be converted, or if it
+                contains such an object.
+        \note  You must call \ref FLValue_Release when finished with the result. */
+    FLValue FLValue_FromNSObject(id);
```
pasin, line 85: Not in exp file.

pasin (commented):
Look great. Found a few small things.

API/fleece/Fleece.h
snej: We have to use standard formats recognized by the compiler, for the compile-time parameter checking to work. `%b` isn't standard.

Fleece/Core/Builder.cc
snej: Oops, fixed. And I added a check for 0x7F (DEL), which counts as a control character.

API/fleece/Fleece+CoreFoundation.h
snej: Added.

pasin (approved):

borrrden:
Is this still valid?  A lot of this stuff seems already present.

pasin (approved):

PR #101, opened 2021-06-10
Fixed crash in alloc_slice::uncopiedNSData()

snej:
The CoreFoundation callbacks needed to gracefully ignore NULL params.
Added a unit test, verified that it reproduced the reported crash, and that it no longer crashes.
Fixes CBL-1988


jayahariv (approved):

jayahariv:
this will be merged to dev branch as well right?

PR #104, opened 2021-07-30
Fixed Android 11 incompatibility with ValueSlot ptr tagging

snej:
The prior size optimization in ValueSlot used the high byte of the
`_pointer` field to distinguish between an actual pointer and an
inline 7-bit Value; it assumed if the upper byte is zero then it's a
pointer.

This isn't true anymore in Android: "Starting in Android 11, for 64-
bit processes, all heap allocations have an implementation defined tag
set in the top byte of the pointer on devices with kernel support for
ARM Top-byte Ignore (TBI)."

Switched to using the _low_ byte as a tag, with a value of 0xFF
denoting inline, and anything else being a pointer. This works because
any Value* we store there must have at least one zero bit in the
LSB; see the comment in ValueSlot.hh for details.

### DIFF
--- a/Fleece/Mutable/ValueSlot.cc
+++ b/Fleece/Mutable/ValueSlot.cc
@@ -37,7 +37,7 @@ namespace fleece { namespace impl {
 
 
     ValueSlot::ValueSlot(Null)
-    :_pointerTag(0xFF)
+    :_tag(kInlineTag)
     {
         _inlineVal[0] = ((kSpecialTag << 4) | kSpecialValueNull);
     }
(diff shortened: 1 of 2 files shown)


snej:
Tested on macOS, needs testing on ARM obviously, esp. with Android 11!

borrrden (approved):
Looks sane to me, but @bmeike has the appropriate hardware for testing.

bmeike (approved):
I just got a clean run of all Android tests, after cherry-picking this commit onto Core master.

I think it is good to go.

Thanks!

snej:
OK, I'll update the Fleece submodule in LiteCore now.

PR #106, opened 2021-08-03
Fixed leak-producing FLSliceResult -> alloc_slice conversions

snej:
The constructor `alloc_slice(FLSliceResult &&)` has been marked `explicit` since
commit 777db2 on Feb 25. I don't know why that change was made. Jianmin just
discovered that our many implicit conversions from an FLSliceResult rvalue to
an alloc_slice are now leaking as a result -- they now call the _lvalue_
reference version, which does not release the FLSliceResult.

I removed `explicit` from the rvalue-ref version and added it to the lvalue-ref
version. This required only two small source changes in LiteCore, and the
LiteCore tests still pass.

(The `explicit slice(const FLSliceResult&)` constructor I added is to fix some
conversions in LiteCore that were apparently using alloc_slice as an
intermediate, and broke when I made the lvalue version explicit.)

### DIFF
--- a/API/fleece/slice.hh
+++ b/API/fleece/slice.hh
@@ -287,6 +287,7 @@ namespace fleece {
         // FLSlice interoperability:
         constexpr slice(const FLSlice &s) noexcept STEPOVER           :pure_slice(s.buf,s.size) { }
         inline explicit operator FLSliceResult () const noexcept;
+        explicit slice(const FLSliceResult &sr) STEPOVER            :pure_slice(sr.buf, sr.size) { }
         slice& operator= (FLHeapSlice s) noexcept           {set(s.buf, s.size); return *this;} // disambiguation
 
 #ifdef SLICE_SUPPORTS_STRING_VIEW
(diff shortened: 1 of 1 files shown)


API/fleece/slice.hh
```diff
@@ -287,6 +287,7 @@ namespace fleece {
         // FLSlice interoperability:
         constexpr slice(const FLSlice &s) noexcept STEPOVER           :pure_slice(s.buf,s.size) { }
         inline explicit operator FLSliceResult () const noexcept;
+        explicit slice(const FLSliceResult &sr)                     :pure_slice(sr.buf, sr.size) { }
```
jianminzhao, line 290: Why does this need to be different from "constexpr slice(const FLSlice &s) noexcept STEPOVER" ?

snej: It doesn't make sense to be `constexpr` because `FLSliceResult` can't exist at compile-time (no heap.)

But it should be STEPOVER (which just keeps debugger Step-In commands from stepping into it.)

snej: * It doesn't need to be constexpr because you can't have FLSliceResult instances at compile-time (no heap!)
* I made it `explicit` because explicit conversions can be dangerous and do unintended things. It's an uncommon conversion, and making it explicit didn't break anything in LiteCore.
* I should probably make it STEPOVER; all that does is stop the debugger from trying to step into it, which is annoying.

jianminzhao (approved):

borrrden (approved):

PR #107, opened 2021-08-20
Do not register scopes of size-2 slice, CBSE-10529

jianminzhao:
Values of size 2 are simple values in that they don't have sub-values. Therefore, they don't provide interesting scope. An exception is the empty dict. We use the empty dict for empty revision bodies of databases and apply technical check, c.f. DatabaseImpl::validateRevisionBody().
We are mostly concerned of the pre-encoded size-2 values, such as Value::kTrueValue, etc. Their address ranges are constant, regardless of the associated scope, causing the failure of the assertion of, "Incompatible duplicate Scope."
However, we should *always* register the scope if this is a Doc.

### DIFF
--- a/Fleece/Core/Doc.cc
+++ b/Fleece/Core/Doc.cc
@@ -110,6 +110,21 @@ namespace fleece { namespace impl {
         Log("Register   (%p ... %p) --> Scope %p, sk=%p [Now %zu]",
             _data.buf, _data.end(), this, _sk.get(), sMemoryMap->size()+1);
 
+        if (!_isDoc && _data.size == 2) {
+            // Values of size 2 are simple values in that they don't have sub-values. Therefore, they don't provide
+            // interesting scope. An exception is the empty dict. We use the empty dict for empty revision bodies
+            // of databases and apply technical check, c.f. DatabaseImpl::validateRevisionBody().
+            // We are mostly concerned of the pre-encoded size-2 values, such as Value::kTrueValue, etc. Their address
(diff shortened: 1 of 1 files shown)


borrrden (approved):
Seems good to me, except I needed the "why does it matter if size == 2" part explained to me

Fleece/Core/Doc.cc
```diff
@@ -121,11 +136,14 @@ namespace fleece { namespace impl {
                 Log("Duplicate  (%p ... %p) --> Scope %p, sk=%p",
                     _data.buf, _data.end(), this, _sk.get());
             } else {
+                static const char* const valueTypeNames[] {"Null", "Boolean", "Number", "String", "Data", "Array", "Dict"};
+                auto type1 = ((const Value*)_data.buf)->type();
+                auto type2 = ((const Value*)existing->_data.buf)->type();
```
snej, line 141: This will not work -- the root Value in encoded Fleece data is (basically) at the end, not the beginning. It only happens to be at the beginning in the degenerate case when the size is 2.

If you want to get the root type you'd have to call `Value::fromData(_data)->type()`, which will parse the data.

snej (changes requested):

Fleece/Core/Doc.cc
jianminzhao: But it will get the right type if the length of the slice is 2, or not? Or perhaps I should try to get the type of general case.

snej (approved):

PR #108, opened 2021-08-20
Fix buffer overflow when decoding base64 string

pasin:
Add one extra bytes to the output buffer as required by the decoder when it reports the ending position for the case that the buffer size is exactly the same size as the output.

CBL-2264

### DIFF
--- a/Fleece/Support/Base64.cc
+++ b/Fleece/Support/Base64.cc
@@ -25,6 +25,7 @@ namespace fleece { namespace base64 {
 
     std::string encode(slice data) {
         std::string str;
+        // 3 bytes -> 4 chars : ceil(data.size / 3) * 4
         size_t strLen = ((data.size + 2) / 3) * 4;
         str.resize(strLen);
         char *dst = &str[0];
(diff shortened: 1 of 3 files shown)


Fleece/Support/Base64.cc
```diff
@@ -39,7 +40,10 @@ namespace fleece { namespace base64 {
 
 
     alloc_slice decode(slice b64) {
-        size_t expectedLen = (b64.size + 3) / 4 * 3;
+        // 4 chars -> 3 bytes : (ceil(b64.size / 4) * 3) + 1
+        // One extra buffer required by the libb64 decoder when reporting ending decoding position
```
jianminzhao, line 44: This is weird as why we need to give an extra byte. And I found there is actually a bug in cdecode.c, line 41, 
--- a/vendor/libb64/cdecode.c
+++ b/vendor/libb64/cdecode.c
@@ -38,7 +38,7 @@ size_t base64_decode_block(const uint8_t* code_in, const size_t length_in, void*
                                if (codechar == code_in+length_in)
                                {
                                        state_in->step = step_a;
-                                       state_in->plainchar = *plainchar;
+                                       state_in->plainchar = 0;
                                        return plainchar - (uint8_t*)plaintext_out;
                                }
                                fragment = base64_decode_value(*codechar++);
This bug wont't affect the correctness, but upset the memory analyzer. It reads the byte beyond the being allocated, but it won't use what's been read. Ideally, we can fix here and make our code more logical.

pasin: That is right. I was not confident to make the change in that area as I'm not sure if it would break anything else. Per chat in slack, as the step at that point is returning the result, the step was at the beginning again. I think the fix is valid. I will make the change.

jianminzhao (approved):

borrrden (approved):

PR #109, opened 2021-08-21
FLDoc: Added get/set associated pointer API

snej:
This provides a reasonably type-safe way to associate a client object with a Fleece `Doc` (`FLDoc`), and thereby with any Value.

For details see the doc-comments in Fleece.h.

This is for use by Couchbase Lite C, for looking up a CBLBlob object given an FLDict in a query result set.

### DIFF
--- a/Fleece/Core/Doc.cc
+++ b/Fleece/Core/Doc.cc
@@ -306,6 +306,20 @@ namespace fleece { namespace impl {
         return RetainedConst<Doc>((const Doc*)scope);
     }
 
+
+    bool Doc::setAssociated(void *pointer, const char *type) {
+        if (_associatedType && type && (_associatedType != type))
+            return false;
+        _associatedPointer = pointer;
+        _associatedType = type;
+        return true;
+    }
+
+    void* Doc::getAssociated(const char *type) const {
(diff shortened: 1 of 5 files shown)


snej:
FYI, this is for CBL-2269.

Fleece/Core/Doc.cc
```diff
@@ -306,6 +306,20 @@ namespace fleece { namespace impl {
...
+    bool Doc::setAssociated(void *pointer, const char *type) {
+        if (_associatedType && type && (_associatedType != type))
+            return false;
+        _associatedPointer = pointer;
+        _associatedType = type;
+        return true;
+    }
+
+    void* Doc::getAssociated(const char *type) const {
+        return (type == _associatedType || type == nullptr) ? _associatedPointer : nullptr;
```
pasin, line 319: If type is nullptr but _associatedType is not nullptr, the _associatedPointer is also returned. Is this expected?

snej: Yeah, I left it as a way to get the pointer no matter what type it is. Not encouraged, but it could be useful for debugging or logging.

pasin (approved):

borrrden (approved):
Is this similar enough to the FLEncoder `setExtraInfo` to consolidate these two method names with each other?

PR #114, opened 2021-08-27
Do not register scopes of size-2 slice, CBSE-10529 (#107)

jianminzhao:
Do not register scopes of size-2 slice, CBSE-10529
Values of size 2 are simple values in that they don't have sub-values. Therefore, they don't provide interesting scope. An exception is the empty dict. We use the empty dict for empty revision bodies of databases and apply technical check, c.f. DatabaseImpl::validateRevisionBody().
We are mostly concerned of the pre-encoded size-2 values, such as Value::kTrueValue, etc. Their address ranges are constant, regardless of the associated scope, causing the failure of the assertion of, "Incompatible duplicate Scope."
However, we should *always* register the scope if this is a Doc.


jianminzhao:
We push the fix to hotfix_android_2.8.7

PR #128, opened 2021-11-03
NumConversion: Improved ParseDouble

snej:
The existing ParseDouble is dumber than ParseInteger, as it doesn't do any error checking. Added a new version that does.


snej:
PS: Not for Lithium -- I'm adding this for future work I'm doing (MQTT)

borrrden (approved):
Don't worry about the Lithium comment, Lithium already has a branch here in fleece so anything in master is already post Lithium

PR #130, opened 2021-11-24
CBL-2591 fixing empty optinal warning by MSVC.

jianminzhao:


borrrden (approved):
As a bonus, you solved the mystery of the GITHUB_ACTIONS hang!

PR #133, opened 2021-12-03
CBL-2615 : Add fleece::RetainedArray and RetainedDict

pasin:
* Added fleece::RetainedArray and RetainedDict in companion with the existing fleece::RetainedValue but specifically for Array and Dict.

* The API and implementation are based on fleece::RetainedValue.

* This is related to CBL-2572 that a CBL-C++ requires to return a retained fleece::Array.

### DIFF
--- a/API/fleece/Mutable.hh
+++ b/API/fleece/Mutable.hh
@@ -249,7 +249,11 @@ namespace fleece {
         }
     };
 
+    // NOTE: The RetainedArray and RetainedDict classes are the copycats of the RetainedValue class
+    // above. Any future changes or bug fixes to the three classes should go together.
 
+    /** Equivalent to Array except that, it holds the Array or MutableArray, and it will retain the
+        underlining FLArray object. */
     class RetainedArray : public Array {
     public:
         RetainedArray()                                 =default;
@@ -283,7 +287,8 @@ namespace fleece {
         }
     };
 
-
+    /** Equivalent to Dict except that, it holds the Dict or MutableDict, and it will retain the
+        underlining FLDict object. */
     class RetainedDict : public Dict {
     public:
         RetainedDict()                                  =default;
--- a/API/fleece/Mutable.hh
+++ b/API/fleece/Mutable.hh
@@ -143,6 +143,7 @@ namespace fleece {
     private:
         MutableArray(FLMutableArray a, bool)     :Array((FLArray)a) {}
         friend class RetainedValue;
+        friend class RetainedArray;
         friend class Array;
     };
 
@@ -207,6 +208,7 @@ namespace fleece {
     private:
         MutableDict(FLMutableDict d, bool)      :Dict((FLDict)d) {}
         friend class RetainedValue;
+        friend class RetainedDict;
         friend class Dict;
     };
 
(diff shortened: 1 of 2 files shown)


borrrden (approved):

jianminzhao:
why cannot use RetainedValue? Isn't Array an Value?

pasin:
This is for a public facing C++ API in CBL-C. If returning RetainedValue, it doesn't indicate that the returned value is an array and users will need to cast or convert it to an array themselves.

The current API signature is `fleece::Array getIndexNames()` (CBL-2572) which has the problem that the underlining fleece array has already been released when the Array is returned.

jianminzhao:
I think the only Array that can be retained is MutableArray. (I may be wrong. @snej please correct/comment) So, the goal here is to demote a MutableArray to a sheer array. The problem here is, Array::~Array() does not release. I feel RetainedArray can be much simpler than the full-fledge one in this PR. 
@snej ?

pasin:
Yes, MutableArray is retaining its underling fleece array. I have chatted with Jens about the possible solution including returning the value as MutableArray but I felt like it's more correct to return an immutable form. Also I tried to match the API (contructors and assignment operators) of the RetainedArray with the RetainedValue. If there is a simpler implementation, please let me know.

jianminzhao:
How about ? 
class RetainedArray: public Array
{
public:
  RetainedArray(FLMutableArray);
  ~RetainedArray();
// if we want to work with RetainedValue,
  operator RetainedValue() const;
}

pasin:
The trimmed down version to support only FLMutableArray instead of FLArray would work as well. However, why do you think we should trim down the API instead of making it being consistent with the RetainedValue?

jianminzhao:
Actually we cannot plug a pure FLArray into it; it will throw at runtime, I think. (If it is inside a Doc, we may retain it, but it will take some extra wiring, I guess).

pasin:
From my chat with @snej, the assumption about the FLArray is that it will either be inside an FLDoc or an FLMutableArray.

pasin:
@snej Could you review the PR and provide your input about the API / implementation?

API/fleece/Mutable.hh
```diff
@@ -247,6 +249,78 @@ namespace fleece {
...
+
+    /** Equivalent to Array except that, it holds the Array or MutableArray, and it will retain the
+        underlining FLArray object. */
+    class RetainedArray : public Array {
+    public:
+        RetainedArray()                                 =default;
+        RetainedArray(FLArray v) noexcept               :Array(FLArray_Retain(v)) { }
+        RetainedArray(const Array &v) noexcept          :Array(FLArray_Retain(v)) { }
+        RetainedArray(RetainedArray &&v) noexcept       :Array(v) {v._val = nullptr;}
+        RetainedArray(const RetainedArray &v) noexcept  :Array(FLArray_Retain(v)) { }
```
snej, line 263: Should be unnecessary? The `(const Array &v)` constructor should cover this.

API/fleece/Mutable.hh
```diff
@@ -247,6 +249,78 @@ namespace fleece {
...
+
+    /** Equivalent to Dict except that, it holds the Dict or MutableDict, and it will retain the
+        underlining FLDict object. */
+    class RetainedDict : public Dict {
+    public:
+        RetainedDict()                                  =default;
+        RetainedDict(FLDict v) noexcept                 :Dict(FLDict_Retain(v)) { }
+        RetainedDict(const Dict &v) noexcept            :Dict(FLDict_Retain(v)) { }
+        RetainedDict(RetainedDict &&v) noexcept         :Dict(v) {v._val = nullptr;}
+        RetainedDict(const RetainedDict &v) noexcept    :Dict(FLDict_Retain(v)) { }
```
snej, line 298: Unnecessary? (See above.)

snej (approved):
Thanks!

jianminzhao (approved):

PR #137, opened 2022-01-13
Add UWP backtrace support

borrrden:
Note: The ability to symbolicate is not available in UWP, so offsets are used instead.  This can be symbolicated using external tools.


Fleece/Support/Backtrace.cc
```diff
@@ -251,9 +252,42 @@ namespace fleece {
...
+                } else {
+                    strcpy(pos, "<unknown module name> ");
+                }
+
+                if(i > 0) {
+                    out << "\r\n"; 
+                }
+
+                out << '\t';
+                out << pos << "+0x" << setw(sizeof(size_t) * 2) << setfill('0') << hex << (uint32_t)((unsigned char *)_addrs[i] - moduleBaseTyped);
```
snej, line 283: The `hex` modifier is persistent, so this will leave the stream writing numbers in hex. Should probably put a `dec` at the end.

(There's another `hex` a few lines down too)

snej (changes requested):

Fleece/Support/Backtrace.cc
borrrden: Oh!  I thought it was a one-off for some reason.  Fixed.

snej (approved):

PR #139, opened 2022-01-22
Fix/format literals

snej:
- In Xcode, enabled -Wformat -Wformat-nonliteral -Wformat-security
- Fixed a couple of places that were insecurely passing non-literalstrings to format args.
- Declared "_throw" etc. as being __printflike to fix warnings.
- Disabled format security around two places where it's warning about things that are weird-but-OK .

### DIFF
--- a/API/fleece/slice.hh
+++ b/API/fleece/slice.hh
@@ -192,7 +192,7 @@ namespace fleece {
         inline bool toCString(char *buf, size_t bufSize) const noexcept;
 
         // FLSlice interoperability:
-        operator FLSlice () const noexcept          {return {buf, size};}
+        constexpr operator FLSlice () const noexcept {return {buf, size};}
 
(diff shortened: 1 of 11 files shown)


snej:
@borrrden : Can we turn these flags on in GCC too? (And their equivalent if any in MSVC)?

borrrden:
Looks like for GCC we can use

```
-Wformat=2

    Enable -Wformat plus additional format checks. Currently equivalent to -Wformat -Wformat-nonliteral -Wformat-security -Wformat-y2k. 
```

Will check on MSVC

borrrden:
MSVC is nonsense, when enabling the warning it gets hit dozens of times by the system headers and so is not very useful.

borrrden (approved):

snej:
FYI, the corresponding LiteCore PR is https://github.com/couchbase/couchbase-lite-core/pull/1363.

PR #140, opened 2022-01-25
Fix warning

Dushistov:
Fix the most dangerous and easy to fix gcc/clang warnings.
The most of them in includes so I get them when I use fleece as external library.
For "strict aliasing violation" I confirm that my fix do not change assembly (on optimization level -O2 | /O2) for clang/gcc/msvc (amd64).

### DIFF
--- a/API/fleece/FLSlice.h
+++ b/API/fleece/FLSlice.h
@@ -188,7 +188,9 @@ static inline void FLSliceResult_Release(FLSliceResult s) FLAPI {
 
 /** Type-casts a FLSliceResult to FLSlice, since C doesn't know it's a subclass. */
 static inline FLSlice FLSliceResult_AsSlice(FLSliceResult sr) {
-    return *(FLSlice*)&sr;
+    FLSlice ret;
+    memcpy(&ret, &sr, sizeof(ret));
+    return ret;
 }
 
 
--- a/API/fleece/Fleece.hh
+++ b/API/fleece/Fleece.hh
@@ -37,6 +37,7 @@ namespace fleece {
     class Value {
     public:
         Value()                                         =default;
+        Value(const Value &) noexcept                   =default;
         Value(FLValue FL_NULLABLE v)                    :_val(v) { }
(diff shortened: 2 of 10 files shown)


borrrden:
Interesting.  The assembly does not change despite introducing two `memcpy` calls into some locations that used to have none?  What is the explanation for this?  How does it get out of those two calls?  Is it smart enough to know what the code is trying to do and reduces it back to the casting or is it dumb and takes the original casting and inserts two `memcpy` ?

Dushistov:
> Interesting. The assembly does not change despite introducing two memcpy calls into some locations that used to have none? 

You can check it via godbolt, here godbolt with sample code: https://godbolt.org/z/zeKs6c1Pz

Almost all C/C++ compiler optimize memcpy if the last argument is constant.
This was the proper solution even ~10-15 years ago, when I fixed such code with UB (undefined behavior):
```
char *p = ...;
reinterpret_cast<unsigned int*>(p);
```
This code cause SIGBUS on Solaris/Sparc (wrong align), and when I replaced it with `memcpy` 
to fix undefined behavior `memcpy` call was optimized away.

borrrden:
Wow, I'm thoroughly impressed that it turns out this way.  I thought of using unions to try to accomplish the same goal but the source is uglier and if this results in no change to the assembly then I'll happily include it.

Dushistov:
>  I thought of using unions to try to accomplish the same goal 

Actually it is impossible to use union in C++ for the same goal. See
https://stackoverflow.com/questions/11373203/accessing-inactive-union-member-and-undefined-behavior
> The confusion is that C explicitly permits type-punning through a union, whereas C++ (c++11) has no such permission.
> 3.8/1 says an object's lifetime ends when its storage is reused. That indicates to me that a non-active member of a union's lifetime has ended because its storage has been reused for the active member. That would mean you're limited in how you use the member

So as I know the only way that valid from C++ standard of view is `memcpy`.

snej (approved):

borrrden (approved):
Thanks!

PR #146, opened 2022-02-11
Split Fleece.h and Fleece.hh into multiple headers

snej:
Fleece.h is now an umbrella header that #includes the new headers:
  FLBase.h, FLCollections.h, FLDeepIterator.h, FLDoc.h, FLEncoder.h,
  FLJSON.h, FLKeyPath.h, FLMutable.h, FLValue.h

There is also FLExpert.h, which is stuff that clients are unlikely to
use, or too dangerous, or semi-deprecated.

On the C++ side I’ve only split out the expert stuff, into Expert.hh.

Also I renamed Base.h to CompilerSupport.h, to keep it from getting
mixed up with FLBase.h (not to mention Base.hh in LiteCore.)


borrrden:
So everything that comes from the umbrella header should be exported as part of CBL for C?

API/fleece/Expert.hh
```diff
@@ -0,0 +1,193 @@
...
+#include "Fleece.hh"
+#endif
+
+#include "FLExpert.h"
+
+FL_ASSUME_NONNULL_BEGIN
+
+namespace fleece {
+
+    //  Rarely-needed or advanced functionality; C++ wrapper around fleece/Obscure.h
```
borrrden, line 26: Wrapped around what?  Should this be FLExpert.h ?

borrrden (changes requested):
Just a comment mistake I think.

API/fleece/Expert.hh
snej: oops, yeah, it was briefly called Obscure.h before I went with "expert".

borrrden (approved):

pasin (approved):

PR #152, opened 2022-03-22
Add a few more headers to the "public" directory

borrrden:
This way we can scoop up a single directory and have all the headers we need to bundle along with a prebuilt Fleece / LiteCore.

:warning: This requires an update to include statements, the following header includes will need to be changed (or overall include path updated):

- function_ref.hh -> fleece/function_ref.hh
- InstanceCounted.hh -> fleece/InstanceCounted.hh
- PlatformCompat.hh -> fleece/PlatformCompat.hh
- RefCounted.hh -> fleece/RefCounted.hh


pasin (approved):
Look good to me. I assume that if one change is missed, the PR validation will not passed as Fleece cannot be built.

PR #153, opened 2022-03-27
Use `std::multiset` for `Scope` memory map

blaugold:
This PR changes the container which stores `Scope` `memEntry`s from a vector to a multiset to improve performance for registering and unregistering `Scope`s.

The complexity of finding `memEntry`'s in a multiset is the same as for a vector (both logarithmic). Insertion and removal, on the other hand, is linear for vectors and logarithmic for multisets.

Note that multisets maintain insertion order for elements that compare as equal and are therefore equivalent to the previously used vector.

[This issue](https://github.com/couchbase/couchbase-lite-C/issues/285) contains more information on the motivation for this change.

### DIFF
--- a/Fleece/Core/Doc.cc
+++ b/Fleece/Core/Doc.cc
@@ -44,12 +44,14 @@ namespace fleece { namespace impl {
         bool operator< (const memEntry &other) const        {return endOfRange < other.endOfRange;}
     };
 
-    // The number of Scopes that can be registered without heap allocation.
-    static const size_t kMemoryMapPrealloc = 10;
-
     using memoryMap = std::multiset<memEntry>;
 
     static memoryMap *sMemoryMap;
+
+    // The maximum number tombstones that will be created in sMemoryMap.
+    // While the number of registered Scopes is below this value, new Scopes can generally be
+    // registered without heap allocation.
+    static const size_t kMemoryMapMaxTombstones = 25;
 
     // The current number of tombstones in sMemoryMap.
     static size_t sMemoryMapTombstones = 0;
@@ -113,12 +115,8 @@ namespace fleece { namespace impl {
             _dataHash = _data.hash();
 #endif
         lock_guard<mutex> lock(sMutex);
-        if (_usuallyFalse(!sMemoryMap)) {
+        if (_usuallyFalse(!sMemoryMap))
             sMemoryMap = new memoryMap;
-            for (size_t i = 0; i < kMemoryMapPrealloc; ++i)
-                sMemoryMap->insert({nullptr, nullptr});
-            sMemoryMapTombstones = kMemoryMapPrealloc;
-        }
         Log("Register   (%p ... %p) --> Scope %p, sk=%p [Now %zu]",
             _data.buf, _data.end(), this, _sk.get(), memEntryCount()+1);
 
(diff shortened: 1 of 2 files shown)


borrrden:
FYI it was not a vector that you replaced, but a smallVector which has different characteristics.  Notably, it will attempt to allocate the first N items on the stack inside of itself rather than performing heap allocations.  I'll ask @snej to take a look at this.

snej:
Honestly, I had forgotten that Scope doesn't use a `map` -- looks like I changed it to a vector in 2019 (dbd0304) because "this performs far less heap allocation and is a lot faster for small numbers of entries as in typical use." Then later in 2019 I changed it to `smallVector` (76c1310)

I can imagine that with thousands of open documents a `map` becomes faster. I just hadn't anticipated that someone in the real world would need that many open documents. If you're doing time-series data (as mentioned in the other issue) it probably makes more sense to store a bunch of events in an array in one document, just because CBL documents are not as lightweight as relational-db rows.

`multimap` isn't necessary. There will never be two entries at the same address, or even that overlap -- see the checks in the `registr` method at line 129.

The problem is that changing back to a `map` will be a performance regression for more common use cases. I don't have specific numbers, but I know that I made this change after running the Instruments tool on a replication task and finding Scope::registr being a hot-spot.

snej:
Two possibilities:
- Switch between a vector to a map at a certain size
- Make the vector scale better by reducing the number of insertions/deletions. A removed entry could be left as a tombstone; when a binary search on insertion hits a tombstone it can reuse that slot. It can garbage-collect by copying into a smaller vector when the number of tombstones grows too large.
- Look around for optimized map implementations (in Folly or Abseil, perhaps, if they don't drag in too many dependencies)

blaugold:
> multimap isn't necessary. There will never be two entries at the same address, or even that overlap -- see the checks in the registr method at line 129.

I don't think that's true, unless I'm misunderstanding what you are saying. [Duplicate docs](https://github.com/couchbase/fleece/blob/37d28cd7b6f52596c9bf0d94213ed7c9a639267d/Tests/ValueTests.cc#L192-L212) will result in two or more entries for the same address range. This seems to be something LiteCore or CBL C is using because the CBL C tests start failing with crashes caused by `RefCounted` releases, when `smallVector` is replaced with a `set` instead of a `multiset`, IIRC. As a side note: I never quite understood how `Doc::containing` can work with duplicate docs.

For alternatives to `smallVector`, I think there is another option. How about using a `multiset` with tombstones to remove the need for heap allocations at small sizes? Nodes can be extracted and inserted without memory allocation. Instead of removing the last N elements, they are turned into tombstones which can be reused the next time an entry needs to be inserted.

snej:
You're right, I forgot that completely identical duplicate scopes are allowed.

snej:
multiset with tombstones sounds good too. Either way, a tombstone can be indicated by setting the `Scope*` to nullptr, I think. Want to implement this?

blaugold:
Sure! I'll update this PR.

blaugold:
I have updated the PR to preallocate and reuse 10 memory map entries.

blaugold:
@snej Friendly ping, to take another look at this PR.

snej:
Dang, sorry. Dropped this on the floor. I'll leave this tab open and try to get to it today!

blaugold:
No problem! It happens sometimes :)

kaptnkoala:
@snej Any news on this?

snej:
Oh man, I suck. At code reviews at least. 😭 Reviewing now...

snej (approved):

blaugold:
😄 Thanks for merging. 🎉

PR #154, opened 2022-04-13
fix some warnings about "cast increases required alignment"

Dushistov:
Fix warnings that produces clang with "-Wcast-align" flags:

```
cast from 'uint8_t *' (aka 'unsigned char *') to 'uint16_t *' (aka 'unsigned short *') increases required alignment from 1 to 2 [-Wcast-align]
ast from 'uint8_t *' (aka 'unsigned char *') to 'uint32_t *' (aka 'unsigned int *') increases required alignment from 1 to 4 [-Wcast-align]
warning: cast from 'fleece::ConcurrentMap::Entry *' to 'uint32_t *' (aka 'unsigned int *') increases required alignment from 2 to 4 [-Wcast-align]
```

The result of cast from pointer with one align to pointer with another align is unspecified
in C++ and undefined behavior in C. So I think it worth to fix,
plus it should be zero cost, because of `memcpy` with fixed size inlined by all modern C++ compilers.

Not fixed:
```
FLSlice.cc:144:16: warning: cast from 'uint8_t *' (aka 'unsigned char *') to 'fleece::sharedBuffer *' increases required alignment from 1 to 4 [-Wcast-align]
        return (sharedBuffer*)((uint8_t*)buf  - offsetof(sharedBuffer, _buf));
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

and 
```
Array.cc:68:21: warning: cast from 'const fleece::impl::Value *' to 'fleece::impl::ValueSlot *' increases required alignment from 1 to 8 [-Wcast-align]
            return ((ValueSlot*)v)->asValue();
                    ^~~~~~~~~~~~~
```

### DIFF
--- a/Fleece/Core/Pointer.hh
+++ b/Fleece/Core/Pointer.hh
@@ -13,6 +13,7 @@
 #pragma once
 #include "Value.hh"
 #include "betterassert.hh"
+#include <cstring>
 
 namespace fleece { namespace impl { namespace internal {
     using namespace fleece::impl;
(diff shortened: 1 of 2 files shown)


borrrden (approved):

borrrden:
Thanks, always good to get rid of these pesky smells!

PR #155, opened 2022-04-13
Create an object library for Fleece with source based exporting

borrrden:
An object library is a prelinked collection of object files that can be consumed by another project and linked in as if they were compiled directly into the consuming project.  This allows some optimizations like compiling once for both static and shared libraries and eliminates the need to keep track of common compiler flags between multiple targets.  

Furthermore, this means that if objects are marked as exported via attribute, the consuming target will honor that attribute.  This is not true with static libraries.


pasin (approved):
1. For CBL-C, as the iOS platform is built using XCode and links to the static libs, we still need to keep the export symbol files?

2. There is a discussion about relocatable object target in this [SO](https://stackoverflow.com/questions/3276474/symbol-hiding-in-static-libraries-built-with-xcode) post. It seems like it's same as the Object lib here?

borrrden:
Yes, until Xcode has a resolution we need to keep at least the apple export files.  I was unable to get it to do what I wanted when I tried to setup Xcode myself.

PR #156, opened 2022-04-15
CBL-2769 : Cleanup compile options between Fleece and LiteCore

jianminzhao:


cmake/platform_win.cmake
```diff
@@ -45,6 +47,29 @@ function(setup_build)
...
+    # Disable the following warnings:
+    #   4068 (unrecognized pragma)
+    #   4244 (converting float to integer)
+    #   4018 (signed / unsigned mismatch)
+    #   4819 (character that cannot be represented in current code page)
+    #   4800 (value forced to bool)
+    #   5105 ("macro expansion producing 'defined' has undefined behavior")
+    # Disable warning about "insecure" C runtime functions (strcpy vs strcpy_s)
+
+    get_all_targets(all_targets)
```
borrrden, line 62: This is a clever trick but we should limit the target we are setting to ones that we have created ourselves and not set our flags on third party code.

borrrden (changes requested):

cmake/platform_win.cmake
jianminzhao: But since we set set the global cmake variables CMAKE_CXX_FLAGS and CMAKE_C_FLAGS. Likes it or not, all sub targets would get them unless they explicitly reset the CMAKE variables. This is the dilemma. If I don't apply it to all sub targets, then those targets that do not reset the CMAKE_CXX_FLAGS will find difference. If I apply to all targets, then those that do reset will find difference.

jianminzhao: After removing the global assignment to CMAKE_C/XX_FLAGS, there are options.
A. keep the current semantics by putting the flags to all targets. This may change the behavior of those sub-targets that reset the flags. This is rare.
B. only ensure that our lite-core targets get those flags, by enumerating those targets.

borrrden (approved):
Yes!  The Windows github check is so much cleaner looking now!

borrrden:
This just needs to update the fleece submodule post merge

PR #162, opened 2022-09-09
CBL-3366 : FLDoc_FromResultData leaks memory

jianminzhao:
It uses an aurgument of FLSliceResult, but fails to release it. The leak is found by valgrind.

### DIFF
--- a/API/fleece/Fleece.hh
+++ b/API/fleece/Fleece.hh
@@ -326,9 +326,8 @@ namespace fleece {
             FLSharedKeys FL_NULLABLE sk =nullptr,
             slice externDest =nullslice) noexcept
         {
-            auto sliceResult = FLSliceResult(std::move(fleeceData));
+            FLSliceResult sliceResult {fleeceData.buf, fleeceData.size};
             _doc = FLDoc_FromResultData(sliceResult, trust, sk, externDest);
-            FLSliceResult_Release(sliceResult);
         }
 
         static inline Doc fromJSON(slice_NONNULL json, FLError* FL_NULLABLE outError = nullptr);
--- a/API/fleece/Fleece.hh
+++ b/API/fleece/Fleece.hh
@@ -325,8 +325,11 @@ namespace fleece {
             FLTrust trust =kFLUntrusted,
             FLSharedKeys FL_NULLABLE sk =nullptr,
             slice externDest =nullslice) noexcept
-        :_doc(FLDoc_FromResultData(FLSliceResult(std::move(fleeceData)), trust, sk, externDest))
-        { }
+        {
+            auto sliceResult = FLSliceResult(std::move(fleeceData));
+            _doc = FLDoc_FromResultData(sliceResult, trust, sk, externDest);
+            FLSliceResult_Release(sliceResult);
+        }
 
         static inline Doc fromJSON(slice_NONNULL json, FLError* FL_NULLABLE outError = nullptr);
 
--- a/API/fleece/Fleece.hh
+++ b/API/fleece/Fleece.hh
@@ -326,6 +326,9 @@ namespace fleece {
             FLSharedKeys FL_NULLABLE sk =nullptr,
             slice externDest =nullslice) noexcept
         {
+            // We construct FLSliceResult the following way to avoid unnecessary
+            // retain. (alloc_slice::operator FLSliceResult()& will apply a retain, which,
+            // if not matched by a release, will lead to memory leak.)
             FLSliceResult sliceResult {fleeceData.buf, fleeceData.size};
             _doc = FLDoc_FromResultData(sliceResult, trust, sk, externDest);
         }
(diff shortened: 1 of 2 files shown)


borrrden (approved):

Fleece/API_Impl/Fleece.cc
```diff
@@ -786,7 +786,9 @@ FLSliceResult FLEncoder_Finish(FLEncoder e, FLError * FL_NULLABLE outError) FLAP
 
 
 FLDoc FLDoc_FromResultData(FLSliceResult data, FLTrust trust, FLSharedKeys FL_NULLABLE sk, FLSlice externData) FLAPI {
-    return retain(new Doc(alloc_slice(data), (Doc::Trust)trust, sk, externData));
+    auto ret = retain(new Doc(alloc_slice(data), (Doc::Trust)trust, sk, externData));
+    FLSliceResult_Release(data);
```
jayahariv, line 790: wait, this might cause trouble in other places right? As per the rule, one who allocates should release the memory. Argument `FLSliceResult` will be passed in from other places, which means, function doesn't owns this. 

https://github.com/couchbase/couchbase-lite-ios/blob/40ff77055a3ba8edeaac71795e1d2dc91c718b22/Objective-C/CBLDocument.mm#L260

jianminzhao: It's a good point. Case in point is this,
`FLDoc FLDoc_FromResultData(FLSliceResult data, FLTrust trust, FLSharedKeys FL_NULLABLE sk, FLSlice externData) FLAPI {
    return retain(new Doc(alloc_slice(data), (Doc::Trust)trust, sk, externData));
}`
in the context of this,
`Doc(alloc_slice fleeceData,
            FLTrust trust =kFLUntrusted,
            FLSharedKeys FL_NULLABLE sk =nullptr,
            slice externDest =nullslice) noexcept
        :_doc(FLDoc_FromResultData(FLSliceResult(std::move(fleeceData)), trust, sk, externDest))
        { }`
Who owns the FLSliceResult as passed in FLDoc_FromResultData() ? In this particular case, the FLSliceResult grabs the ownership from alloc_slice and becomes the sole owner. It's bound to leak if function, FLDoc_FromResultData(), does not release it. Doctrine: pass the ownership to the callee.
If we insist on this doctrine that caller owns FLSliceResult, then, the code would have to flow as
`FLSliceResult flSliceResult = FLSliceResult(<move or not> fleeceData);
 _doc(FLDoc_FromResultData(flSliceResult, trust, sk, externDest);
FLSliceResult_Release(flSliceResult);`

borrrden: This is not the rule for FLSliceResult.  It is exactly the opposite, when you pass along an FLSliceResult you are passing ownership as well, otherwise it would just be FLSlice.

jianminzhao: As a corollary, the callee should release it before exiting?

borrrden: That is not a requirement, the callee can do whatever they wish as long as it is eventually released.

jianminzhao: Of course, it can keep the ownership (ref-count). If the callee is done with it, it must release it. Is this correct?

borrrden: Yes, that is what ownership means.  The owner must either release it, or hand it off to another owner.

jianminzhao: With regard to Jay's example, https://github.com/couchbase/couchbase-lite-ios/blob/40ff77055a3ba8edeaac71795e1d2dc91c718b22/Objective-C/CBLDocument.mm#L260, the caller passed "body" to FLDoc_FromResultData without pre-retaining it. After it is done with the SliceResult, the callee is supposed to release it, as like my previous commit. Then that code will crash: the caller's "body" will have its ref-count decreased.

borrrden: I've lost track of what you are talking about.  So what if he doesn't retain it?  That's not a requirement and if he did it would be a leak unless he then released it again.

jianminzhao: This is a redacted code snippet,
`   FLSliceResult body = FLEncoder_Finish(encoder, &flErr);  <------------------------------- (1)
    ....
        FLDoc doc = FLDoc_FromResultData(body, kFLTrusted, db.sharedKeys, nullslice);  <---- (2)
  ....
    return body;`        <------------------------------------------------------------------------(3)
In line (1), we obtain body. Suppose it's ref-count equals n.
In line (2), we pass it to FLDoc_FromResultData. Note that we do not retain it, i.e., ref-count == n. This function, FLDoc_FromResultData, actually keeps a copy in the form of alloc_slice. The ref-count is boosted to n+1. The function cannot know whether the caller means to keep a ref-count or not. In this case, the caller wants it and return it to its caller, at line (3). Therefore, the function cannot release it. In the case of the memory leak as Valgrind detected, the caller simply passes it to FLDoc_FromResultData and the ref-count must be released by the function, or the memory will be leaked.

borrrden: As discussed offline, I was under mistaken preconditions and the correct way was decided.

API/fleece/Fleece.hh
```diff
@@ -326,9 +326,8 @@ namespace fleece {
             FLSharedKeys FL_NULLABLE sk =nullptr,
             slice externDest =nullslice) noexcept
         {
-            auto sliceResult = FLSliceResult(std::move(fleeceData));
+            FLSliceResult sliceResult {fleeceData.buf, fleeceData.size};
```
borrrden, line 329: This probably deserves a comment on why it is doing things like this (i.e. to avoid over-retaining).

PR #164, opened 2022-09-20
CBL-3686: add a new retain function, retainUnlessUnique, to RefCounted

jianminzhao:
It retains the RefCounted only if the ref-count is greater than 1. Otherwise, it returns false without retaining.

### DIFF
--- a/API/fleece/RefCounted.hh
+++ b/API/fleece/RefCounted.hh
@@ -39,6 +39,8 @@ namespace fleece {
     private:
         template <typename T>
         friend T* retain(T*) noexcept;
+        template <typename T>
+        friend bool retainUnlessUnique(T*) noexcept;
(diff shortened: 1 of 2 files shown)


snej (commented):
It occurs to me that you can do the same thing without having to modify RefCounted. Wouldn't this be identical:
```c++
retain(x);
if (x.retainCount() == 2) release(x);
```

jianminzhao:
What a perspicacious observation: I think this should work. If the ref-count goes back to two after I release here, it would indicate a bug similar to that count goes 0 then back up.
I think I can make do without this PR, except sounds little esoteric.

jianminzhao:
Taking @jens trick and abandon this PR, so that we don't have to add a new non-standard API to RefCounted

PR #166, opened 2022-09-29
Issues discovered through fuzzing

blaugold:
I've been experimenting with testing Fleece with fuzzed Fleece data. 

In the process, I had to expand the validation of untrusted Fleece data. It seems to be quite complete now. Fuzzing for an extended amount of time does not produce new assert failures or crashes.

I found and fixed the following issues:

- An assert is not handling comparing two `nan` floats.
- `ValueSlot::copyValue` does not handle integers and binary data.
- An `undefined` value in a `Dict` causes an incorrect count in deep mutable copies.
  The problem is that in [this line](https://github.com/couchbase/fleece/blob/master/Fleece/Mutable/HeapDict.cc#L118) `_source->get(key)` returns `nullptr` for a `undefined` value, so `ValueSlot::setting` is not able to distinguish between a new key and a key with an `undefined` value.
  I'm not sure if my fix is the right approach, though.

The fuzzing function I used is relatively simple. With a more advanced function (e.g. also handling shared keys) it might be possible to discover more issues.

```c++
#include <cstdlib>

#include "fleece/Fleece.hh"
#include "fleece/Mutable.hh"
#include "fleece/Expert.hh"

static void IterateValue(fleece::Value& value) {
    fleece::DeepIterator copyIter(value);
    for (; copyIter; ++copyIter) {}
}

extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
    fleece::alloc_slice slice(Data, Size);
    fleece::Doc doc(slice);

    if (doc) {
        auto root = doc.root();
        IterateValue(root);

        switch (root.type()) {
            case kFLArray: {
                auto copy = root.asArray().mutableCopy(kFLDeepCopyImmutables);
                IterateValue(copy);
                break;
            }
            case kFLDict: {
                auto copy = root.asDict().mutableCopy(kFLDeepCopyImmutables);
                IterateValue(copy);
                break;
            }
            default:
                return 0;
        }
    }

    return 0;
}
```

Fuzzing with LLVM's libFuzzer is quite straightforward and could be automated in CI.

### DIFF
--- a/Fleece/API_Impl/FLSlice.cc
+++ b/Fleece/API_Impl/FLSlice.cc
@@ -41,15 +41,28 @@ FL_ASSUME_NONNULL_BEGIN
 
 __hot
 bool FLSlice_Equal(FLSlice a, FLSlice b) noexcept {
-    return a.size == b.size && FLMemCmp(a.buf, b.buf, a.size) == 0;
+    if (a.size == b.size) {
+        if (a.size == 0)
+            // Check wether both slices are the nullslice or empty slices.
+            return (a.buf == nullptr) == (b.buf == nullptr);
+        else
+            return FLMemCmp(a.buf, b.buf, a.size) == 0;
+    }
+    return false;
 }
-
 
 __hot
 int FLSlice_Compare(FLSlice a, FLSlice b) noexcept {
     // Optimized for speed, not simplicity!
     if (a.size == b.size)
-        return FLMemCmp(a.buf, b.buf, a.size);
+        if (a.size == 0) {
+            // Sort the nullsice before an empty slice.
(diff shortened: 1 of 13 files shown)


blaugold:
Tow more issues:

- Fix the count of a dict that inherits all of its entries from the parent.
- Distinguish a `nullslice` from an empty slice`. The two are not the same, but `FLSlice_Equal` and `FLSlice_Compare` treats them as if they are.

blaugold:
@snej @borrrden I have created separate PRs for fixes for the discovered issues. Please take a look.

- #170
- #171 
- #172
- #173
- #174

PR #168, opened 2022-10-04
Don't cache `Dict::key::_numericKey` when `SharedKeys` is in transaction

blaugold:
A transaction might be rolled back, potentially leaving an invalid shared key in a `Dict::key`.

Related issue: https://github.com/couchbase/couchbase-lite-C/issues/287

### DIFF
--- a/Fleece/Core/SharedKeys.hh
+++ b/Fleece/Core/SharedKeys.hh
@@ -106,6 +106,8 @@ namespace fleece { namespace impl {
 
         bool isUnknownKey(int key) const FLPURE;
 
+        bool isInTransaction() const FLPURE             {return _inTransaction;}
+
         virtual bool refresh()                          {return false;}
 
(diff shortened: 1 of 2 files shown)


Fleece/Core/SharedKeys.hh
```diff
@@ -106,6 +106,8 @@ namespace fleece { namespace impl {
 
         bool isUnknownKey(int key) const FLPURE;
 
+        virtual bool isInTransaction() const FLPURE     {return _inTransaction;}
```
snej, line 109: Remove `virtual` since this isn't overridden anywhere.

snej (changes requested):
Just one little fix...

Fleece/Core/SharedKeys.hh
blaugold: Oh yeah, I initial thought I would have to override it in `PersistentSharedKeys`. Done.

PR #170, opened 2022-11-06
Fix comparing `nullslice` and empty slice

blaugold:
`FLSlice_Equal` and `FLSlice_Compare` handled the `nullslice` and an empty slice as beeing equal.

This can cause incorrect behaviour in edge cases where the distinction matters. The added tests contains such a case and did fail before this change.

NB: This was discovered through fuzzing (https://github.com/couchbase/fleece/pull/166).


snej:
I honestly don't recall if this was intentional or not. I worry that changing this behavior could cause regressions in LiteCore or Couchbase Lite.

snej:
I copied the test case in here, "API Mutable copy of Dict with empty string shared key", into the current master version of Fleece and ran it; it passes.

PR #172, opened 2022-11-06
Handle binary and int values in `ValueSlot::copyValue`

blaugold:
With this change all types of values that might not be inlinable into `ValueSlot` are handled in `ValueSlot::copyValue`.

NB: This was discovered through fuzzing (#166).


snej (approved):
Good catch! I must've forgotten binary entirely, and mis-assumed ints would always be small enough to fit inline.

PR #174, opened 2022-11-06
Fix assert to handle `nan`

blaugold:
Fixes the assert in `ValueSlot::set` to handle `nan` values.

NB: This was discovered through fuzzing (#166).


snej:
Not sure it's a good idea to support NaNs; they can't be exported to JSON, and they have weird behavior overall. Did you find a situation where they were produced somehow without explicitly feeding one to the API?

PR #178, opened 2023-05-26
To allow Xcode 14 to build hydrogen with deprecated sprintf

jianminzhao:
Until we migrate all sprintf's to snprint's, we need to disable GCC_WARN_ABOUT_DEPRECATED_FUNCTIONS .


pasin (changes requested):
Can this be set in xcconfig file?

jianminzhao:
I am not familiar with it. I did it in Xcode. Let me try  it.

pasin (changes requested):
Can you update the value here instead of creating a new entry?

https://github.com/couchbase/fleece/pull/178/files#diff-b7306dd4c870cee3de73d109a1676a6892fb0aea30b451c8063fb901580e42e3R38

pasin (approved):

PR #180, opened 2023-06-08
Add build configuration Test_EE based on Debug_EE

jianminzhao:
This is motivated by CBL 4546

### DIFF
--- a/Fleece.xcodeproj/project.pbxproj
+++ b/Fleece.xcodeproj/project.pbxproj
@@ -1673,6 +1673,59 @@
 			};
 			name = Release;
 		};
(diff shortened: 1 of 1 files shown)


jianminzhao:
Note: the new configuration, Test_EE, is created in Xcode UI, create configuration based on Debug_EE
I need it because in LiteCore, I created Test_EE configuration, also based on Debug_EE, to which only a new DEFINE, LITECORE_CPPTEST, is added. I only need to link to fleece built from Debug_EE, but it does not seem that Xcode would allow it. Xcode will build Test_EE of dependent libs.

PR #183, opened 2023-07-07
Add a new Xcode config, Test_CE. It is a duplicate of Debug. We need …

jianminzhao:
…it for 'LiteCore C++ Tests'/Test_CE. C.F. CBL 4687

### DIFF
--- a/Fleece.xcodeproj/project.pbxproj
+++ b/Fleece.xcodeproj/project.pbxproj
@@ -1673,6 +1673,57 @@
 			};
 			name = Release;
 		};
+		D6200C0A2A5746B200790C0F /* Test_CE */ = {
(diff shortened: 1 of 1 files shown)


callumbirks (approved):
Not opposed to this at all, but I just wonder what is the motivation for it?

borrrden (commented):
What about CMake?

jianminzhao:
For CMake, it is clear. I can a define to the CMake (pseudo) library, LiteCoreUnitTesting, which is the exactly the same as LiteCoreObjects except for an additional DEFINE, and CppTest executable links to LiteCoreUnitTesting rather than LiteCoreObjects. For Xcode, I have to add DEFINEs per build configuration.

borrrden (approved):

PR #185, opened 2023-07-14
CBL-4680: Change jsonsl submodule from snej to couchbase deps

jianminzhao:


borrrden (approved):
Don't forget to also change the manifest!

PR #186, opened 2023-07-19
Some improvements/fixes needed for my VersionVector work

snej:
Sorry if it's awkward having these in one PR, but it seemed a lot less work than creating five separate branches with their own PRs.

It's best to review each commit separately. They're all pretty small.


borrrden (approved):
This was actually very readable and this format is fine, especially if each commit describes a piece of the overall work in a tidy way like this.

PR #187, opened 2023-08-16
CBL-4824: Move fixes for 3.1.2 from `next/3.1` to `release/3.1`

callumbirks:
Looks like Core release branch requires Fleece submodule points to matching release branch.
This is merged from master up to the point required for CBL-4824, which is the Xcode 14.3 fixes that allow Core to build.

### DIFF
--- a/Xcode/xcconfigs/FleeceTest.xcconfig
+++ b/Xcode/xcconfigs/FleeceTest.xcconfig
@@ -12,7 +12,6 @@
 //  the file licenses/APL2.txt.
 //
 
-// Workaround for Catch 2 requiring C++ features not available until macOS 10.12
 MACOSX_DEPLOYMENT_TARGET            = 10.14
 
 CLANG_WARN__EXIT_TIME_DESTRUCTORS = NO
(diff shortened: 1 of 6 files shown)


borrrden (changes requested):
There are beryllium tickets in the commits here, I think those will need to be changed.

velicuvlad:
[ea80355](https://github.com/couchbase/fleece/pull/187/commits/ea80355fa41036d50f1b13c438b56363faf7115b) has to be taken out. We moved to 11 and 10.14 for 3.1 as well

callumbirks:
Yeah I think I did this the wrong way. Going to redo

callumbirks:
Okay so redid this with a cherry-pick of just CBL-4471 ([8e10e3d](https://github.com/couchbase/fleece/commits/8e10e3d0243a1204d2bda589ea13f93404b0c6d2)) which is the only commit in Fleece which needs to brought forward for the 3.1.2 fixes that were sat in the next branch

Xcode/xcconfigs/FleeceTest.xcconfig
```diff
@@ -13,7 +13,7 @@
 //
 
 // Workaround for Catch 2 requiring C++ features not available until macOS 10.12
```
jianminzhao, line 15: Is the comment still valid?

borrrden (approved):

Xcode/xcconfigs/FleeceTest.xcconfig
callumbirks: Probably not, no. I can remove it

PR #193, opened 2023-11-06
fix: bump 50 level parsing limitation to 100

admosity:
~~This PR bumps the 50 level parsing limit found in the parser up to `UINT_MAX`.~~

This PR bumps the 50 level parsing limit found in the parser up to 100.

### DIFF
--- a/Fleece/Core/JSONConverter.cc
+++ b/Fleece/Core/JSONConverter.cc
@@ -32,7 +32,7 @@ namespace fleece { namespace impl {
 
     JSONConverter::JSONConverter(Encoder &e) noexcept
     :_encoder(e),
-     _jsn(jsonsl_new(UINT_MAX)),      // never returns nullptr, according to source code
+     _jsn(jsonsl_new(100)),      // never returns nullptr, according to source code
      _jsonError(JSONSL_ERROR_SUCCESS),
      _errorPos(0)
     {
--- a/Fleece/Core/JSONConverter.cc
+++ b/Fleece/Core/JSONConverter.cc
@@ -32,7 +32,7 @@ namespace fleece { namespace impl {
 
     JSONConverter::JSONConverter(Encoder &e) noexcept
     :_encoder(e),
-     _jsn(jsonsl_new(50)),      // never returns nullptr, according to source code
+     _jsn(jsonsl_new(UINT_MAX)),      // never returns nullptr, according to source code
      _jsonError(JSONSL_ERROR_SUCCESS),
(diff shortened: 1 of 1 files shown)


caraie (approved):

caiello-disney (approved):

Fleece/Core/JSONConverter.cc
```diff
@@ -32,7 +32,7 @@ namespace fleece { namespace impl {
 
     JSONConverter::JSONConverter(Encoder &e) noexcept
     :_encoder(e),
-     _jsn(jsonsl_new(50)),      // never returns nullptr, according to source code
+     _jsn(jsonsl_new(UINT_MAX)),      // never returns nullptr, according to source code
```
snej, line 35: Did you happen to notice what this does inside jsonsl? From jsonsl.c line 111:
```c
            calloc(1, sizeof (*jsn) +
                    ( (nlevels-1) * sizeof (struct jsonsl_state_st) )
```
This patch will cause jsonsl to allocate a huge amount of memory -- 4 billion times the size of a struct that's about 50 bytes large, i.e. around 200GB.

The OS kernel may or may not commit all that address space. If it doesn't, things are OK. If it does, that's an immediate out-of-memory error. I suspect iOS in particular will have the latter behavior since it doesn't fully use virtual memory.

snej:
We could consider bumping the 50 to a larger number like 100, but I'm inclined to say that it's not reasonable for an app to create documents that are nested 50+ levels deep. If it can do that, wouldn't there be a scenario that creates even deeper nesting that will break at whatever deeper limit we set?

Fleece/Core/JSONConverter.cc
admosity: Thank you! It's an honor to get told by the creator of the ubiquitous chat bubble 🙇🏻 

I apologize for the oversight- it has been a long time since I have used c++ and had made some assumptions about the implementation. This is definitely a no no.

admosity:
Thanks for getting back and I agree with this. We had unexpectedly hit this error downstream (through the couchbase sync gateway). We have a Server Driven UI use case where we are generating json that creates deeply nested structures. We are actively working on something to properly split these documents.

I believe that 100 would be a nice bump as to allow the splitting process to not create a high level of fragmented documents.

snej (approved):

jianminzhao:
We will increase the depth limit to 100, [CBL-5311](https://issues.couchbase.com/browse/CBL-5311)

PR #194, opened 2023-11-09
Do not capture the backtrace for OutOfRange exception.

jianminzhao:
As the iterator moves beyond the end of the array or dict, the implementation will throw FleeceException with error code OutOfRange. This is necessary to handle the damaged iterator. However, we don't need to capture the backtrace in this case, which is very expensive as shown in a customer's performance profile, c.f. CBL 5044.


callumbirks (approved):

Fleece/Support/FleeceException.cc
```diff
@@ -48,8 +48,11 @@ namespace fleece {
     :std::runtime_error(what)
     ,code(code_)
     ,err_no(errno_)
-    ,backtrace(Backtrace::capture(2))
-    { }
+    {
+        if (code_ != OutOfRange) {
```
pasin, line 52: I think it's worth to have a comment here why OurOfRange is ignored.

pasin (approved):

PR #195, opened 2023-11-18
Add a separate property, isCacheable, to SharedKeys.

jianminzhao:
If cacheable, Dict::get(key &keyToFind) will cache the string key to integer key and enhance the performance of the ensuing gets. This is good if the doucment is not changed.

For PersistentSharedKey, the library manages the property: it is cacheable only if it is not in a transaction when Dict::get is called.

For SharedKeys, the client manages it. If it anticipates the sharedKeys may outlive the underlying document changes, it should call SharedKeys::disableCaching().


snej (changes requested):
`LITECORE_CPPTEST` isn't an accurate name because this isn't the LiteCore project. How about `FLEECE_TEST`?

jianminzhao:
The reason of LITECORE_CPPTEST is because the automatic tests are run from LiteCore. The cmake project of CppTests lumps all tests from fleece with the LiteCore's tests.

Tests/ValueTests.cc
```diff
@@ -262,7 +262,7 @@ namespace fleece {
 
         caughtException = false;
         // ++iter will throw if already at the end.
-        // OutOfRange exception should contain the backtrace.
+        // OutOfRange exception should not contain the backtrace.
```
jianminzhao, line 265: This is a mistake. Should be "... not contain the backtrace."

snej (approved):

PR #198, opened 2024-01-03
CBL-59: Supporting date parsing functions

callumbirks:
Rebased feature/parse_date_parts onto master.

### DIFF
--- a/Fleece/Support/ParseDate.cc
+++ b/Fleece/Support/ParseDate.cc
@@ -93,23 +93,6 @@ typedef int64_t sqlite3_int64;
 
 
 /*
- ** A structure for holding a single date and time.
- */
-typedef struct DateTime DateTime;
-struct DateTime {
-    sqlite3_int64 iJD; /* The julian day number times 86400000 */
-    int Y, M, D;       /* Year, month, and day */
-    int h, m;          /* Hour and minutes */
-    int tz;            /* Timezone offset in minutes */
-    double s;          /* Seconds */
-    char validYMD;     /* True (1) if Y,M,D are valid */
-    char validHMS;     /* True (1) if h,m,s are valid */
-    char validJD;      /* True (1) if iJD is valid */
-    char validTZ;      /* True (1) if tz is valid */
-};
-
-
-/*
  ** Convert zDate into one or more integers.  Additional arguments
  ** come in groups of 5 as follows:
  **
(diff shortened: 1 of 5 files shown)


callumbirks:
The functions exposed / added here are required to support the date/time functions being added to Core / N1QL

callumbirks:
Seems like some functions from the PRD aren't present in any existing branches. (Specifically, `str_to_tz`, and other tz functions). I think I'll need to add a new dependency for that (`date/tz`). We already have `date/date.h` as a Fleece dependency, we just need the timezone extensions.

Fleece/Support/ParseDate.cc
```diff
@@ -329,124 +309,294 @@ static void inject_local_tz(DateTime* p)
...
+                  [](unsigned char c) { return std::tolower(c); });
+        const auto entry = dateComponentMap.find(componentStr);
+        if ( entry == dateComponentMap.end() ) { return kDateComponentInvalid; }
+
+        return entry->second;
     }
 
-    slice FormatISO8601Date(char buf[], int64_t time, bool asUTC) {
-        if (time == kInvalidDate) {
+    slice FormatISO8601Date(char buf[], int64_t timestamp, bool asUTC, const DateTime* format) {
```
jianminzhao, line 473: Why do we need format? Does it mean that the caller may want to discard the YMD or HMS ?

jianminzhao:
Please add some tests

callumbirks:
> Please add some tests

There are lots of tests in the accompanying LiteCore PR (blocked on this one), which test these functions, but through the interface of N1QL

Fleece/Support/ParseDate.cc
callumbirks: Yeah, per the PRD, we must support format other than ISO8601. We provide this in N1QL by allowing a DateTime parameter where the caller can pass a DateTime that looks like the format they want. i.e. if they pass "11:11:11" we would return them a datetime that only has HMS

callumbirks:
@jianminzhao There is a test already (`SupportTests.cc:334`) which converts timestamp to string, and back again. That process involves most/all of the functions added/modified here

pasin (approved):
From @callumbirks, we have tests that well cover the all the parser logic in this PR.

jianminzhao (approved):

PR #199, opened 2024-01-15
CBL-59: Fix for core

callumbirks:
This fixes 2 issues:
- Apparently I forgot to run some of the tests in Core, so an incorrect cast which was causing overflow is now fixed.
- Core has stricter compilation warnings and errors than Fleece, so an issue which was caught by Core's build (`result` var unused), was not caught by the Fleece PR / build.

For the second issue, this could be avoided if we make Fleece have the same warnings / error settings as Core.

### DIFF
--- a/Fleece/Support/ParseDate.cc
+++ b/Fleece/Support/ParseDate.cc
@@ -431,10 +431,11 @@ namespace fleece {
     DateTime FromMillis(int64_t timestamp) {
         // Split out the milliseconds from the timestamp:
         time_t secs{timestamp / 1000};
-        int    millis = (int)timestamp % 1000;
+        int    millis = (int)(timestamp % 1000);
 
         struct tm  timebuf {};
         struct tm* result = gmtime_r(&secs, &timebuf);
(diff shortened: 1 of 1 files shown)


Fleece/Support/ParseDate.cc
```diff
@@ -431,10 +431,11 @@ namespace fleece {
     DateTime FromMillis(int64_t timestamp) {
         // Split out the milliseconds from the timestamp:
         time_t secs{timestamp / 1000};
-        int    millis = (int)timestamp % 1000;
+        int    millis = (int)(timestamp % 1000);
```
jianminzhao, line 434: Do you need to narrow_cast it?

borrrden: If it is % 1000 it is guaranteed not to overflow or underflow.

jianminzhao: max of int64_t is roughly 2^63, and max of int (assuming 32 bit is roughly 2^31.

jianminzhao (approved):

PR #201, opened 2024-01-23
CBL-59: Windows fix (Removing `gmtime_r`)

callumbirks:


callumbirks:
It turns out that Windows' implementation of `gmtime` does not match the C spec at all, they do not accept negative epoch values (so no dates before 1970). Thanks a lot to @borrrden I've managed to replace the usages of gmtime with modern C++ equivalents. 
This, I am 90% sure, is actually the last Fleece PR for this ticket.
I am going to point Core to this branch though, just to make sure.

jianminzhao:
Wow, a lot of changes.

jianminzhao (approved):

PR #213, opened 2024-03-11
CBL-5438: DateTime standard format parser

callumbirks:
A proper parser for DateTime formats, rather than just parsing format string as a DateTime.
This has benefits such as additional flexibility in future, but also allows us to keep consistent with the strange "almost-ISO" date format we shipped with CBL 3.1, because keeping format as a `DateTime` does not allow us to optionally use colon or no-colon for Timezone offset (i.e. `+0500` / `+05:00`).
I added in parsing for a "subset" of https://howardhinnant.github.io/date/date.html#to_stream_formatting, but it's not necessary and if anyone thinks we don't need it, happy to take it out. Think it's a lot better than the weird "1111-11-11T11:11......` format we copied from Server though.

### DIFF
--- a/Fleece/Support/DateFormat.cc
+++ b/Fleece/Support/DateFormat.cc
@@ -13,7 +13,7 @@ namespace fleece {
     // YYYY-MM-DDThh:mm:ssTZD
     DateFormat DateFormat::kISO8601 = DateFormat{YMD{Year{}, Month{}, Day{}, YMD::Separator::Hyphen},
                                                  Separator::T,
-                                                 HMS{Hours{}, Minutes{}, Seconds{}, HMS::Separator::Colon},
+                                                 HMS{Hours{}, Minutes{}, Seconds{}, Millis{}, HMS::Separator::Colon},
                                                  {Timezone{}}};
 
     /** This parses a subset of the formatting tokens from "date.h", found 
@@ -321,7 +321,7 @@ namespace fleece {
         }
     }
 
-    slice DateFormat::format(char buf[], int64_t timestamp, minutes tzoffset, std::optional<DateFormat> fmt) {
+    slice DateFormat::format(char buf[], const int64_t timestamp, minutes tzoffset, std::optional<DateFormat> fmt) {
         if ( timestamp == kInvalidDate ) {
             *buf = 0;
             return nullslice;
@@ -341,7 +341,7 @@ namespace fleece {
         if ( f.hms.has_value() ) {
             if ( f.ymd.has_value() ) { stream << (char)f.separator.value(); }
 
-            if ( f.hms.value().millis.has_value() ) {
+            if ( f.hms.value().millis.has_value() && timestamp % 1000 ) {
                 stream << date::format("%T", tm);
             } else {
                 const auto secs = duration_cast<seconds>(millis);
@@ -360,7 +360,7 @@ namespace fleece {
         }
 
         const std::string res = stream.str();
-        std::strncpy(buf, res.c_str(), res.length());
+        strncpy(buf, res.c_str(), res.length());
 
         return {buf, res.length()};
     }
--- a/Fleece/Support/DateFormat.cc
+++ b/Fleece/Support/DateFormat.cc
@@ -9,15 +9,176 @@ namespace fleece {
 
     using namespace std::chrono;
 
+    // The default format
     // YYYY-MM-DDThh:mm:ssTZD
-    DateFormat DateFormat::kISO8601 = DateFormat{YMD{{}, {}, {}, YMD::Separator::Hyphen},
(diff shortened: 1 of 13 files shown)


callumbirks:
See Core PR https://github.com/couchbase/couchbase-lite-core/pull/1977

callumbirks:
First issue: Unsure whether CBL3.1 included milliseconds by default or not. I thought it didn't, so I removed milliseconds from the default format (found at `DateFormat::kISO8601`). However, this has broken some Fleece tests that don't expect to lose the milliseconds precision.

borrrden:
Are there tests for this somewhere?  Could they be added to the Fleece tests?

Also I'm kind of unsure if this qualifies as feature creep or not since this is getting out of the scope of a binary encoding.  @snej Do you have an opinion on this?  This functionality needs to be somewhere accessible to Fleece obviously but is it ok to live next to fleece itself?

callumbirks:
I added a couple tests in Core. Probably does need more.

And I'm not sure. Previously we parse date format strings as DateTime, which restricts us severely. For example, we must use no-colon by default in Timezone specifier, but using DateTime as format we can't offer colon delimited Timezone. So as far as the parser goes, I don't see that the scope is any bigger than it already was (although it may have already had scope creep). However, the %-format is not strictly necessary (although much nicer).

borrrden:
Can those Core tests be moved here in some fashion and use the underlying classes directly?

callumbirks:
That would be a good idea... Having to run a Core PR pointing here to make sure nothing's broken over here isn't the best way to write software....

callumbirks:
@borrrden I've added some DateFormat tests. See `SupportTests.cc`.

Fleece/Support/DateFormat.cc
```diff
@@ -0,0 +1,431 @@
...
+        if ( formatString.size < 2 ) return {};
+
+        // - YMD
+
+        auto ymd = kISO8601.ymd;
+
+        // %F == %Y-%m-%d
+        if ( formatString[1] == 'F' ) {
+            // ISO YMD, so we don't need to change `ymd`
+            formatString.setStart(formatString.begin() + 2);
```
snej, line 49: `slice_istream` would be pretty useful here, making the code simpler & safer

Fleece/Support/DateFormat.hh
```diff
@@ -0,0 +1,127 @@
...
+        enum class Month : uint8_t {};
+        enum class Day : uint8_t {};
+        enum class Hours : uint8_t {};
+        enum class Minutes : uint8_t {};
+        enum class Seconds : uint8_t {};
+        enum class Millis : uint8_t {};
+        enum class Timezone : uint8_t { NoColon, Colon };
+        enum class Separator : char { Space = ' ', T = 'T' };
+
+        struct YMD {
```
snej, line 51: It looks as though the only parts of YMD and HMS that are used by this PR are the separators. I understand wanting room to grow, but all the work of defining types and fields and structs that aren't used yet seems too much ... It seems like each of these structs could be reduced to the Separator enum for now, and it wouldn't be difficult to refactor them back to structs in the future if/when we need it.

snej (changes requested):
Top-level comment is that this is functionality only needed by CBL N1QL functions, not Fleece itself, so I'd rather the code were put in LiteCore.

borrrden:
I believe the counterpoint to "this is only used in LiteCore" is "we were using some of this for a specific function in here which writes a datetime to a stream".  In that case at least some of this needs to be accessible here which makes breaking it up somewhat of a pain.

borrrden (commented):
Jens points out some nice feedback so I'll +1 the code parts that he mentioned.

callumbirks:
Closing. Moved `DateFormat` to Core. See https://github.com/couchbase/couchbase-lite-core/pull/1977

callumbirks:
@borrrden Regarding
> some of this needs to be accessible here
For now I'm just going along with the duplicated code. Jens said he expected some duplicated code around this, and would rather this not be in Fleece. I definitely agree it doesn't really belong in Fleece, but I think the only solution which doesn't involve duplicated code would be adding a whole new dependency. Which I feel would be more pain than it's worth. It would involve ripping out basically all Date-related functionality from Fleece, and requiring it depend on this dependency.

Fleece/Support/DateFormat.hh
```diff
@@ -0,0 +1,127 @@
...
+
+            Hours                  hours;
+            Minutes                minutes;
+            std::optional<Seconds> seconds;
+            std::optional<Millis>  millis;
+            Separator              separator;
+        };
+
+        // 1111-11-11T11:11:11(Z)
+        DateFormat(YMD ymd, Separator separator, HMS hms, const std::optional<Timezone> tz = {})
```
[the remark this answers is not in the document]
callumbirks: But, it is essentially a variable declaration. Which (IMO) should be const if it isn't modified in the function.

PR #217, opened 2024-04-08
Explicit convert UINT64_MAX to unsigned

jianminzhao:
In Jenkins' build machine (MacOS), we found it failed to convert UINT64_MAX to unsigned with @() after we upgraded MACOSX_DEVELOPMENT_TARGET to 12 from 10.14.

A test is falied because of it. Fix it by explicit cast.

### DIFF
--- a/Tests/ObjCTests.mm
+++ b/Tests/ObjCTests.mm
@@ -62,8 +62,7 @@ TEST_CASE("Obj-C Ints", "[Encoder]") {
     checkIt(@123456789,     "123456789");
     checkIt(@123456789012,  "123456789012");
     checkIt(@(INT64_MAX),   "9223372036854775807");
-    // checkIt(@(UINT64_MAX),  "18446744073709551615");
-    // The above may fail to turn UINT64_MAX to unsigned. In the following,
+    // It was found that @() may not turn UINT64_MAX to unsigned. In the following,
     // we apply an explicit conversion
     checkIt(@([@UINT64_MAX unsignedLongLongValue]), "18446744073709551615");
 }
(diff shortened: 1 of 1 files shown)


Tests/ObjCTests.mm
```diff
@@ -62,7 +62,10 @@ static void checkIt(id obj, const char* json) {
     checkIt(@123456789,     "123456789");
     checkIt(@123456789012,  "123456789012");
     checkIt(@(INT64_MAX),   "9223372036854775807");
-    checkIt(@(UINT64_MAX),  "18446744073709551615");
+    // checkIt(@(UINT64_MAX),  "18446744073709551615");
```
pasin, line 65: Can we remove this line?

pasin (approved):

Tests/ObjCTests.mm
jianminzhao: okay. I was waving between whether to remove it.

Tests/ObjCTests.mm
```diff
@@ -62,7 +62,9 @@ static void checkIt(id obj, const char* json) {
     checkIt(@123456789,     "123456789");
     checkIt(@123456789012,  "123456789012");
     checkIt(@(INT64_MAX),   "9223372036854775807");
-    checkIt(@(UINT64_MAX),  "18446744073709551615");
+    // It was found that @() may not turn UINT64_MAX to unsigned. In the following,
+    // we apply an explicit conversion
+    checkIt(@([@UINT64_MAX unsignedLongLongValue]), "18446744073709551615");
```
snej, line 67: This is unnecessarily convoluted. You create an NSNumber, turn it back into a native number, then create another NSNumber from that.
It would be much simpler to write `[NSNumber numberWithUnsignedLongLong: UINT64_MAX]`.
Or else `@((unsigned long long)UINT64_MAX)`

jianminzhao: Thanks for the advice. This is test code and won't be in the product. I will change when ported the the master branch.

PR #222, opened 2024-06-03
CBL-5686: Fix GCC 'nonull' warnings

callumbirks:
Many of the warnings that nonnull GCC spews out seem to come from Fleece. 
Looks like there's a TODO item to replace NONNULL with FL_NONULL, so I've done that. 
Hopefully that addresses the GCC warnings.


callumbirks:
Looks like CLion messed with the formatting a little. I can revert it, if desired

jianminzhao (approved):

PR #230, opened 2024-06-11
Fix FLEncoder_GetErrorMessage call after FLEncoder_Finish [#229]

### LINKED ISSUE #229: FLEncoder_GetErrorMessage always returns NULL after FLEncoder_Finish
Blake was having trouble getting an error message from a Fleece encoder error, and while inspecting the code I noticed that
- FLEncoder_Finish clears the error code
- FLEncoder_GetErrorMessage returns NULL if the error code is 0
Oops...

I decided the best fix was to take that check out of FLEncoder_GetErrorMessage, and instead return NULL only if the error string is empty. FLEncoder_Finish doesn't clear the error string, so it'll still be there afterward.

snej:
Ensure you can still get the error message from an encoder after calling Finish on it.
Fixes #229

### DIFF
--- a/Fleece/API_Impl/Fleece.cc
+++ b/Fleece/API_Impl/Fleece.cc
@@ -719,7 +719,7 @@ FLError FLEncoder_GetError(FLEncoder e) FLAPI {
 }
 
 const char* FL_NULLABLE FLEncoder_GetErrorMessage(FLEncoder e) FLAPI {
-    return e->hasError() ? e->errorMessage.c_str() : nullptr;
+    return e->errorMessage.empty() ? nullptr : e->errorMessage.c_str();
 }
 
 void FLEncoder_SetExtraInfo(FLEncoder e, void* FL_NULLABLE info) FLAPI {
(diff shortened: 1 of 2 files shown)


Fleece/API_Impl/Fleece.cc
```diff
@@ -719,7 +719,7 @@ FLError FLEncoder_GetError(FLEncoder e) FLAPI {
 }
 
 const char* FL_NULLABLE FLEncoder_GetErrorMessage(FLEncoder e) FLAPI {
-    return e->hasError() ? e->errorMessage.c_str() : nullptr;
+    return e->errorMessage.empty() ? nullptr : e->errorMessage.c_str();
```
jianminzhao, line 722: It appears to me that we are taking advantage of the unsynchronized pair (error_code, error_message) after reset() (we only reset with failure). What's purpose to clear error code but not clear the error message? Or put it another way, why do we have to clear the error code, or can we keep both error code and error message after reset()? 
Is the encoder still usable after reset()?

snej: The code implies the encoder is still usable after reset. I don't know if anything is reusing encoders -- I could check our code, but devs using CBL-C also use FLEncoder. So I didn't want to break that functionality.

It's necessary to clear the error code or reusing the encoder will fail immediately. Leaving the message alone seemed safe since no one's going to look at the error message until after there's an error.

jianminzhao (approved):

PR #232, opened 2024-06-13
Merged blaugold's WASM branch, plus fixes

snej:
Cherry-picked @blaugold 's branch that adds WASM support via emscripten.
Made a few minor changes to get it to build again.
Added a Makefile to simplify (all types of) builds.


Fleece/Support/ParseDate.cc
```diff
@@ -558,6 +558,8 @@ namespace fleece {
         auto offset = seconds(-s);
 #elif defined(__DARWIN_UNIX03) || defined(__ANDROID__) || defined(_XOPEN_SOURCE) || defined(_SVID_SOURCE)
         auto offset = seconds(-timezone);
+#elif defined(__EMSCRIPTEN__)
+        auto offset = seconds(timezone);
```
borrrden, line 562: Is this negated on purpose?  Does ECMAScript somehow change the functionality here?

borrrden:
Also none of this builds at all on Windows, is that known?

borrrden:
It also appears to not build on Linux unless a specific version of emscripten is required?  The final link fails possibly because of an unrecognized argument but somewhat unclear and too long to paste here.

Fleece/Support/ParseDate.cc
blaugold: It was necessary to make the tests pass, but it seems like a bug in Emscripten which implements all these UNIX APIs. Not sure what the current behavior is, though.

Iceburgino:
Are there any updates on this?

snej:
I think it would be OK to merge this, since it causes no real changes in "normal" supported builds. The conflicts are probably easy to resolve.

PR #239, opened 2024-09-13
More compiler warnings, and sanitizers, in CMake builds

snej:
Added new CMake options: 

- `FLEECE_WARNINGS_HARDCORE` enables _all_ compiler warnings except for a few dozen that are too onerous, plus `-Werror`.
- `FLEECE_SANITIZE` enables the Clang Address and UB sanitizers.

Currently these are only supported in Clang builds, and ignored with other compilers.

Also made source changes to fix some of the newly-enabled warnings.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -45,12 +45,12 @@ jobs:
     - name: Configure CMake
       if: runner.os != 'Windows'
       working-directory: ${{github.workspace}}/build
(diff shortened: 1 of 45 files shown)


borrrden:
This is probably going to cause more warnings on non clang platforms due to unknown pragmas (clang diagnostic) but it's really just a drop in the bucket when it comes time to actually tackle that and we may well switch to clang everywhere first.

borrrden (approved):

PR #240, opened 2024-10-03
Updated Catch2 to v3.7.1, from v2.13.8

snej:
- catch.hpp is now just a stub that includes catch_amalgamated.hpp. I did this to avoid having to change every file that #includes it.
- Catch is now split into header and implementation, which speeds up build times. Added catch_amalgamated.cpp to targets.
- In CatchHelper.hh I changed the way N_WAY_TEST_CASE_METHOD works; it used to be a kludge but now uses the GENERATE() facility. As a benefit, it now works properly with SECTIONs -- it used to be that an n-way test ignored sections when the test option was != 0.
- I had to reimplement our CaseListReporter from scratch because it's no longer possible to subclass the built-in reporters, and also because the reporter API changed. The only feasible way to keep the nice features of ConsoleReporter was to copy the source code, rename the class and modify that :-p

**NOTE:** LiteCore PR https://github.com/couchbase/couchbase-lite-core/pull/2155 depends on this one, and without it, this PR will break LiteCore.


borrrden:
Eventually we should switch to use the recommended way of doing this, which is via ExternalProject_Add

borrrden (approved):

PR #244, opened 2024-11-20
CBL-6485 : Define LIFETIMEBOUND only for C++

pasin:
* Moved the definition of LIFETIMEBOUND to be next to FLAPI.
* Aligned formatting with the other sections by moving # to the front of the lines.

### DIFF
--- a/API/fleece/CompilerSupport.h
+++ b/API/fleece/CompilerSupport.h
@@ -269,23 +269,23 @@
 #endif
 
 #ifdef __cplusplus
-    #define FLAPI noexcept
-#else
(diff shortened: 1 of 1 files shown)


snej (changes requested):
Actually LIFETIMEBOUND should move to PlatformCompat.hh since it's only meant for C++ code.

pasin:
If moving LIFETIMEBOUND to PlatformCompat.hh, where do we define an empty one for C?

borrrden (approved):

snej (approved):

PR #253, opened 2025-01-17
fix calling base class private method

Dushistov:
because of bugs gcc/clang accept this before,
but now clang 19.x gives compile time error,
because of "untack" method is private:

https://godbolt.org/z/hd9n1o3br

### DIFF
--- a/API/fleece/InstanceCounted.hh
+++ b/API/fleece/InstanceCounted.hh
@@ -48,9 +48,9 @@ namespace fleece {
 
     protected:
         InstanceCounted(size_t offset)              {track(offset);}
+        void untrack() const;
     private:
         void track(size_t offset =0) const;
-        void untrack() const;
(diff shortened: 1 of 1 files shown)


borrrden:
This...is so wild.  It's probably been like this for years, and MSVC still compiles it.  Was this some sort of universal misinterpretation of the C++ spec of something?

Dushistov:
> This...is so wild. It's probably been like this for years, and MSVC still compiles it. Was this some sort of universal 
> misinterpretation of the C++ spec of something?

gcc also still compiles this just fine. There is for example another example https://github.com/llvm/llvm-project/issues/16784 , where gcc warn about private access, while clang compile just fine.

As I understand, the C++ standard tells about this: `no diagnostic required`.
Therefore, the compiler may or may not report an error if the template has not been instantiated.
Another story if template was used in code, then all C++ compilers  have to reports errors.

snej (approved):

PR #258, opened 2025-04-04
Fix incorrectly private method in InstanceCounted.

callumbirks:

PR #259, opened 2025-04-10
Merge master into 3.2

callumbirks:
As mentioned in #258 now might be a good time to merge master into 3.2


callumbirks:
Hmm Litecore 3.2 can't be built with this branch. I imagine some changes need to happen there too, so we'll have to hold off until after release.

borrrden:
In general this is the REVERSE of what we want to do.  We want to only merge forward (i.e. 3.2 -> master) not backwards because newer versions are very likely to contain work that is invalid fro earlier ones.

snej:
I agree with Jim. Closing.

PR #262, opened 2025-04-25
Improvements from the P2P work

snej:
Best to look at each of the 5 commits individually:

### [smallVector: added contains, addUnique, remove](https://github.com/couchbase/fleece/pull/262/commits/378abaa0d63bcedbbe995f7b0895043f07ec6d53)

Simply adds three utility methods. These are useful for using a smallVector as a lightweight set (for small sizes obviously.)

### [RefCounted: Added retained_cast](https://github.com/couchbase/fleece/pull/262/commits/3d1ed61e3744ceb4a4cb0d9c726d6f578a02e2ec)

`retained_cast<T>(U)` is equivalent to `dynamic_cast` except its parameter and result are Retained values not raw pointers.

### [CMake: Added -Wno-unknown-warning-option](https://github.com/couchbase/fleece/pull/262/commits/d2ad82b96583ef3dde185b73ca26a36e1310ec09)

Fixes a build error in Clang 16. [I've actually made this a separate PR, #263, since it's broken the master branch.]

### [slice: Improved use of LIFETIMEBOUND](https://github.com/couchbase/fleece/pull/262/commits/8cfd106abdc1ee6d167d0cd2b535a0541b5e360f)

Helps detect some additional dangling-pointer errors, by adding the LIFETIMEBOUND attribute for conversions from alloc_slice to slice. This tells Clang that the resulting slice's lifetime must not exceed the original alloc_slice's.

There are some other changes that are just to fix some annoying compiler errors triggered by the above.

### [RefCounted: Added non-nullable Retained, AKA Ref](https://github.com/couchbase/fleece/pull/262/commits/fae3f3e1c115d69ec8c6e9369e0a2be0cde4ed0d)

OK, this one is cool! I added a second template parameter to `Retained` so it can be nullable or non-nullable; it's now `Retained<class T, Nullability N>`. The second parameter defaults to `MaybeNull` so normal use of Retained works as before, but now we also have `Retained<T,Nonnull>` -- aliased as `Ref<T>`.

`Ref<T>` is just like `Retained<T>` except it cannot hold nullptr.
- It's very useful in API function parameters as an indication that null isn't legal; and in return values as a promise that the result is never null.
- It has no default constructor; it has to be constructed with a value.
- Its methods use FL_NONNULL so UBSan will detect violations at runtime (and sometimes at compile time.)
- Even without UBSan, putting a null pointer in one causes a failure (NPE) at runtime.
- You can't directly convert a Retained<T> to a Ref<T>; you have to call `asRef()`, which throws illegal_argument if the value is null. (I couldn't think of a better name for this. Something like "IPromiseThisIsNotNull".)
- Ref<T> is able to skip most null checks when calling retain() and release(), making the code slightly smaller & faster.

As a side benefit, I found that the `RetainedConst` class was never necessary; `RetainedConst<T>` is equivalent to `Retained<const T>`.

### DIFF
--- a/API/fleece/RefCounted.hh
+++ b/API/fleece/RefCounted.hh
@@ -274,6 +274,13 @@ namespace fleece {
     }
 
 
+    /** `retained_cast<T>(...)` is like `dynamic_cast` but on pointers wrapped in Retained. */
+    template <class T, class U>
+    Retained<T> retained_cast(Retained<U> const& r) {
+        return dynamic_cast<T*>(r.get());
+    }
+
+
     /** Extracts the pointer from a Retained. It must later be released via `release`.
         This is used in bridging functions that return a direct pointer for a C API. */
     template <typename REFCOUNTED>
--- a/API/fleece/RefCounted.hh
+++ b/API/fleece/RefCounted.hh
@@ -13,9 +13,15 @@
 #pragma once
 #include "fleece/PlatformCompat.hh"
 #include <atomic>
+#include <concepts>
+#include <stdexcept>
 #include <utility>
 
+FL_ASSUME_NONNULL_BEGIN
+
 namespace fleece {
+
+    enum Nullability {NonNull, MaybeNull};
 
     /** Simple thread-safe ref-counting implementation.
         `RefCounted` objects should be managed by \ref Retained smart-pointers:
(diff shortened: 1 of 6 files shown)


jianminzhao (approved):

jianminzhao:
Learn a lot!

borrrden (approved):

PR #272, opened 2025-10-23
Implemented AtomicRetained / AtomicRef

snej:
`AtomicRetained` is a fully thread-safe version of `Retained` that supports concurrent gets and sets.
 It's a lot slower than `Retained`, so use it only when necessary: in situations where it will be called on multiple threads.


snej:
I've updated Replicator to use AtomicRetained in its `SubReplicator` struct, to fix CBL-8457. Replicator unit tests pass.

borrrden:
I think you put the wrong ticket number into the comment?

borrrden (approved):
Looks good to me but it needs to go into the release/4.1 and release/3.4 branches as well

PR #287, opened 2026-03-20
Major reliability improvement for logging stack traces

borrrden:
This took many days of effort and so it's involved a lot with various stack unwinding and address retrieval mechanisms so I'll try to provide a high level summary here.  Firstly to avoid a giant spaghetti of ifdef I broke out the implementation across several files that only get compiled on certain platforms:

- Backtrace+capture-darwin.cc: Apple specific logic for getting a symbol from a frame in a stack trace.  This uses standard `backtrace_symbols` and `dladdr` since the implementation on Apple is much more enhanced than the glibc linux counterpart (Most likely the compiler is more friendly about it when it emits machine code).  
- Backtrace+capture-linux.cc: Linux specific logic for getting a symbol from a frame in a stack trace.  This makes use of the GCC provided libbacktrace.a (it's present even when using clang because by default clang still uses GCC for its runtime stuff in c++).  This library is very attuned to specifically getting frames on Linux even when frame pointers are omitted and handles all the gnarly unwinding stuff safely (I found when trying to implement register based stack crawling that I often got it wrong and caused an SIGSEGV while trying to do it)
- Backtrace+capture-posix.cc: Other stuff common to all POSIX environments like unmangling and actual printing.  In this we need to be very careful since my plan is to handle signals here as well.  We can only use [async signal safe functions](https://www.man7.org/linux/man-pages/man7/signal-safety.7.html).  In particular this means no streams, no printf, and no unmangling either.
- Backtrace+capture-win32.cc: Windows specific logic for getting a symbol from a frame in a stack trace.  Printing is not restricted since the way that Windows faults already moves to an entirely separate context in which it is safe to do just about anything.  This uses the `StackWalk64` API.
- Backtrace+signals-posix.cc: POSIX signal handlers!  RAII style class that will register signal handlers and write out a stack trace, and then pass on control to the next handler.
- Backtrace+signals-win32.cc: Win32 exception handlers!  Windows doesn't really have the equivalent of POSIX signals.  It has a few basic ones but that's at.  Most faults result in one of the various types of exception handlers (this is an overloaded term, since Windows was using the term "exception" way before it had any meaning for c++.  These "exceptions" can happen even in C.  They are basically signals by another name except the way the handlers are invoked is more safe).  
- Backtrace.cc: This now encapsulates any logic that operates one level above all of this logically.  It uses the primitives defines in the files above.  

Note that since we cannot use ostreams in signal handlers, I created a new crash log in the logs directory that a crash will be written to upon a signal.  The existing logs are all buried inside of ofstream objects.


snej (dismissed):

Fleece/Support/Backtrace+capture-posix.cc
```diff
@@ -0,0 +1,207 @@
...
+            {"(anonymous namespace)", "(anon)"},
+            {"std::__1::", "std::"},
+            {"std::__cxx11::", "std::"},
+            {"std::basic_string<char, std::char_traits<char>, std::allocator<char> >", "std::string"},
+    };
+
+    // If any of these strings occur in a backtrace, suppress further frames.
+    static constexpr const char* kTerminalFunctions[] = {
+            "_C_A_T_C_H____T_E_S_T_",
+            "Catch::TestInvokerAsFunction::invoke() const",
```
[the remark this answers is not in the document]
borrrden: I have not.  Is there a method you used to determine them in the first place?

Fleece/Support/Backtrace.hh
```diff
@@ -18,19 +18,48 @@
...
+                                                                void* context = nullptr);
+
+        static int raw_capture(void** buffer, int max, void* context = nullptr);
 
         /// Captures a backtrace, unless maxFrames is zero.
         /// @param skipFrames  Number of frames to skip at top of stack
         /// @param maxFrames  Maximum number of frames to capture
-        explicit Backtrace(unsigned skipFrames =0, unsigned maxFrames =50);
+        /// @param context     If non-null, a platform-specific context to capture from
+        /// @param from       If non-null, the address to start from instead of the current frame
```
jianminzhao: Where is "from" in the signature?
As it is, it can be linked to Core without changes in Core, right? 
How platform passes the context to Backtrace?

borrrden: The "platform" doesn't pass anything, it's all handled via the various Backtrace capture classes I've written.  I think "from" here is a dangling comment from something I removed.

jianminzhao (approved):

PR #297, opened 2026-05-27
Fixed rare buffer overflow with Encoder::snip()

snej:
Encoder::snip() allocates 2 bytes for a Value, but the Value constructor actually initializes all 4 bytes. If the allocation is at the very end of a heap block it'll overwrite the end.

In the case of CBSE-22711 the buffer is the Writer's internal buffer so the overwrite clobbered some following state of the Encoder, causing it to crash on the next use.

Fixes CBL-8363

## Notes on the fix

I changed the Value constructor to only write to the first 2 bytes (i.e. a 'narrow' Value) which was its original intended behavior.

IIRC, the constructor originally did this but I altered it at some point to initialize the entire 4-item `_byte` array so that it could be made `constexpr` (otherwise the compiler complains that the object contains uninitialized data.) We still need the constexpr constructor, but I've differentiated its signature by adding parameters for the other 2 byte values.

### DIFF
--- a/Fleece/Core/Array.hh
+++ b/Fleece/Core/Array.hh
@@ -59,7 +59,7 @@ namespace fleece { namespace impl {
 
         inline iterator begin() const noexcept;
 
-        constexpr Array()  :Value(internal::kArrayTag, 0, 0) { }
+        constexpr Array()  :Value(internal::kArrayTag, 0, 0, 0, 0) { }
 
     protected:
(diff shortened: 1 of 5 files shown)


borrrden (approved):
This test is dependent on the current layout of the Encoder as its written, but address sanitizer will fail on this test as well so it's not as important as it seems.

PR #308, opened 2026-06-20
CBL-8494: Use bit 1 instead of the high bit for AtomicWrapper busy flag

borrrden:
AtomicWrapper's spinlock stored its "busy" flag in the high bit of the pointer (bit 63 on 64-bit). On 64-bit Android, Bionic's tagged pointers (TBI) place an allocator tag in the top byte of every heap pointer, so bit 63 is frequently already set. getAndLock() then reads the slot as permanently busy and spins forever, wedging the owning thread with no crash or ANR. The high bit is also a real address bit on 32-bit ABIs.

Move the flag to bit 1 (value 2). RefCounted has an atomic<int32_t> member, so it is always >= 4-aligned and bits 0 and 1 are guaranteed zero in any stored pointer on every ABI. Bit 0 stays reserved for Fleece's mutable-Value low-bit tag. getAndLock() already returns the bit-cleared value, so no dereference-site masking is needed.

### DIFF
--- a/Fleece/Support/RefCounted.cc
+++ b/Fleece/Support/RefCounted.cc
@@ -135,8 +135,11 @@ namespace fleece {
 
     namespace internal {
         /// Tag bit that's added to `_ref` while accessing it.
-        /// We can't use the low bit (1) because mutable Fleece Values already use that as a tag.
-        static constexpr uintptr_t kBusyMask = uintptr_t(1) << (8 * sizeof(uintptr_t) - 1);
+        /// We use bit 1 (value 2): RefCounted has an `atomic<int32_t>` member, so it's always
+        /// >= 4-aligned, guaranteeing bits 0 and 1 are zero in any stored pointer on every ABI.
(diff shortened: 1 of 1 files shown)


snej:
I'm not sure bit 1 will work for mutable Fleece values. The ‘Value‘ is really an internal field of a heap-allocated HeapValue struct, withs a 1-byte pad before it to give it an odd address. 

The HeapValue will be at least 8-byte aligned in 64-bit, but I don't remember whether the offset of the pad in the whole struct is a multiple of 4

snej:
...ok, it appears bit 1 is available. A HeapValue has a vtable ptr, a 32-bit refcount, a 1-byte pad, then the Value. So the offset of the Value is 13, i.e. 1101. If we assume at least 4-byte alignment of heap allocations, bit 1 is always 0.

snej (approved):

snej:
PS: Happy Father's Day, Jim!

borrrden:
> PS: Happy Father's Day, Jim!

To you as well!

